PR #48, opened 2021-05-24
fix URL

cmungall:
when I used the proposed URL it forces to me to fork then I end up with a blank file https://github.com/cmungall/bioregistry/edit/main/src/bioregistry/data/bioregistry.json


cthoyt:
Ahh thank you. GitHub is sneaky and changes this stuff all the time. It definitely worked before!

PR #54, opened 2021-05-25
Add semweb context

### LINKED ISSUE #52: Create a collection and JSON-LD context for semantic web

cthoyt:
Closes #52
CC @cmungall note there are a few discrepancies with https://github.com/prefixcommons/biocontext/blob/master/registry/semweb_context.jsonld. 

- `idot` entry is fixed
- `dc` is not given as a duplicate for `dc.terms`
- Some of the other URLs changed as well because they're automatically retrieved from the bioregistry

### DIFF
--- a/src/bioregistry/data/collections.json
+++ b/src/bioregistry/data/collections.json
@@ -26,5 +26,29 @@
       "emdb",
       "bmrb"
     ]
+  },
+  "0000002": {
+    "name": "Semantic Web Context",
+    "contextName": "semweb",
+    "description": "Resources used in the semantic web, based on https://github.com/prefixcommons/biocontext/blob/master/registry/semweb_context.yaml.",
+    "author": {
+      "name": "Charles Tapley Hoyt",
+      "orcid": "0000-0003-4423-4370"
+    },
+    "resources": [
+      "ac",
+      "rdf",
+      "rdfs",
+      "owl",
+      "xsd",
+      "dc.terms",
+      "faldo",
+      "foaf",
+      "oa",
+      "idot",
+      "void",
+      "prov",
+      "dcat"
+    ]
   }
 }
--- a/src/bioregistry/data/collections.json
+++ b/src/bioregistry/data/collections.json
@@ -47,6 +47,7 @@
       "foaf",
       "idot",
       "oa",
+      "oboinowl",
       "owl",
       "prov",
       "rdf",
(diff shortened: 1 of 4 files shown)


src/bioregistry/data/collections.json
```diff
@@ -28,5 +28,32 @@
...
+        "name": "Charles Tapley Hoyt",
+        "orcid": "0000-0003-4423-4370"
+      }
+    ],
+    "contextName": "semweb",
+    "description": "Resources used in the semantic web, based on https://github.com/prefixcommons/biocontext/blob/master/registry/semweb_context.yaml.",
+    "name": "Semantic Web Context",
+    "resources": [
+      "ac",
+      "dc.terms",
```
cmungall, line 44: while I do like the idea of sub-namespacing in some situations, it's much more conventional to use `dcterms` as the prefix

cmungall:
I think this is good, some comments

this PR introduces the concept of collections, which seems quite useful

Although I initially chose 'semweb' as the grouping in biocontexts, I am not so sure it's a coherent grouping right now. It encompasses domain-specific schemas such as faldo that follow semweb "norms" (e.g. use of rdfs as a schema language), as well as core w3 vocabularies. I don't have a great alternative suggestion. If things can be sliced and diced in different ways in future that would be great.

I am not sure I understand all the changes but go ahead and merge anyway

cmungall:
Also we may want to think about this in the context of the relationship of bioregistry to http://prefix.cc/ the namespace lookup for rdf developers

cmungall:
http://prefix.cc/context

src/bioregistry/data/collections.json
cthoyt: Addressed in 646c4bd

cthoyt:
> I think this is good, some comments
> 
> this PR introduces the concept of collections, which seems quite useful
> 
> Although I initially chose 'semweb' as the grouping in biocontexts, I am not so sure it's a coherent grouping right now. It encompasses domain-specific schemas such as faldo that follow semweb "norms" (e.g. use of rdfs as a schema language), as well as core w3 vocabularies. I don't have a great alternative suggestion. If things can be sliced and diced in different ways in future that would be great.
> 
> I am not sure I understand all the changes but go ahead and merge anyway

The idea of collections is quite general - anyone can make one and add an arbitrary description about what it is. There's currently only one example at https://bioregistry.io/collection/, and it was me making all of the information in a review machine readable.

This PR decouples the process of defining the scope of the contexts and the generation, so I think this will be nice for later.

I've also seen prefix.cc and they might already be on my todo list on GitHub, thanks for the reminder

PR #58, opened 2021-05-26
Update documentation pages

cthoyt:
This PR reorganizes the previous "documentation" pages for the summary and download into several pages. It also adds a new page on "sustainability" which includes a much more detailed description of how the app is managed as well as allusions to some of the discussions we've been having on the OBO Community slack.

@bgyori it would be nice if you could take a look at the text. Probably easiest to deploy the app yourself and read from the browser with:

```shell
$ git clone git@github.com:bioregistry/bioregistry.git
$ git checkout update-meta
$ pip install -e .
$ bioregistry web
```

### DIFF
--- a/src/bioregistry/app/templates/access.html
+++ b/src/bioregistry/app/templates/access.html
@@ -1,0 +1,117 @@
+{% extends "prose.html" %}
+
+{% block title %}Bioregistry Programmatic Access{% endblock %}
+
+{% block styles %}
+    {{ super() }}
+    <link href="https://cdn.jsdelivr.net/npm/prismjs@1.23.0/themes/prism.css" rel="stylesheet"/>
+{% endblock %}
+
+{% block scripts %}
+    {{ super() }}
+    <script src="https://cdn.jsdelivr.net/npm/prismjs@1.23.0/prism.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/prismjs@1.23.0/plugins/autoloader/prism-autoloader.min.js"></script>
+{% endblock %}
+
+{% block content %}
+    <h2>API Usage</h2>
+    <p>
+        The Bioregistry web application is built on <a href="https://flask.palletsprojects.com">Flask</a>
+        as a thin wrapper around the <a href="https://github.com/bioregistry/bioregistry"><code>bioregistry</code></a>
+        Python package. It exposes several endpoints for accessing the registry, metaregistry, collections, and search
+        functionality for which <a href="https://swagger.io/">Swagger</a> API documentation is automatically generated
+        by <a href="https://github.com/flasgger/flasgger">Flasgger</a>.
+    </p>
+    <p>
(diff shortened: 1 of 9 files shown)


bgyori (commented):
Great, thanks! For me, the new dropdown seems to have transparent background so it overlaps with the text behind it (on Chrome):
image

I pushed a few typo fixes in the summary.

src/bioregistry/app/templates/access.html
```diff
@@ -0,0 +1,117 @@
+{% extends "prose.html" %}
+
+{% block title %}Bioregistry Programmatic Access{% endblock %}
```
bgyori, line 3: Access -> Usage?

src/bioregistry/app/templates/base.html
```diff
@@ -52,11 +52,19 @@
...
-                <a class="nav-link" href="{{ url_for("download") }}">Downloads</a>
+
+
+            <li class="nav-item active dropdown">
+                <a class="nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                    About
+                </a>
+                <div class="dropdown-menu" aria-labelledby="navbarDropdown">
+                    <a class="dropdown-item" href="{{ url_for("summary") }}">Summary</a>
+                    <a class="dropdown-item" href="{{ url_for("download") }}">Downloads</a>
```
bgyori, line 63: I think Downloads could warrant its own heading instead of being part of the "About" section.

src/bioregistry/app/templates/collections.html
```diff
@@ -1,7 +1,7 @@
 {% extends "base.html" %}
 {% import "macros.html" as utils %}
 
-{% block title %}Bioregistry - Colletions{% endblock %}
+{% block title %}Bioregistry - Collections{% endblock %}
```
bgyori, line 4: It would be nice to add a sentence or two on this page about what a "collection" is.

bgyori: On that note, a simple sentence on top of the Registry and Metaregistry pages would also be nice to describe what those are directly on the page.

src/bioregistry/app/templates/base.html
```diff
@@ -52,11 +52,19 @@
...
+                    <a class="dropdown-item" href="{{ url_for("access") }}">Programmatic Access</a>
+                    <a class="dropdown-item" href="{{ url_for("sustainability") }}">Sustainability</a>
```
bgyori, line 65: Maybe "Deployment and sustainability" would be a better - though longer - heading for this section.

src/bioregistry/app/templates/access.html
cthoyt: done in e399847

src/bioregistry/app/templates/base.html
cthoyt: It was actually like that before, I think you're right to revert it. done in e399847

PR #87, opened 2021-06-07
Add additional OBO Foundry prefixes

cthoyt:
While the OBO Foundry doesn't list UMLS, SNOMED-CT, or ENSEMBL as ontologies, they do have a place in the typical usage of prefix maps that accompany usage of OBO Foundry ontologies. Therefore, this PR adds them to the OBO Foundry prefix map. 

All entries in the OBO Foundry are keyed with lowercase names. However, during use, they are typically uppercased, except when there's an explicit `preferredPrefix` entry to show how mixed case might get used, such as for FBbt.

Therefore, in this PR, the keys are also kept lowercase. This PR also adds an extraneous `preferredPrefix` that just uses full caps for demonstration purposes. I also explicitly ran the export command for the OBO Foundry JSON-LD context file, but this is actually generated automatically on the nightly build.

CC @matentzn


cthoyt:
This was not what Nico had in mind, I'm going to close this. See #89 instead

PR #107, opened 2021-07-22
Add override URL for ena.embl

### LINKED ISSUE #106: ena.embl broken link
Links here are broken http://bioregistry.io/registry/ena.embl

I think this is caused by the double space:

https://www.ebi.ac.uk/ena//ena/browser/view/BN000065

this works

https://www.ebi.ac.uk/ena/browser/view/BN000065

(may as well change to https)

and for other types too e.g. projects
https://www.ebi.ac.uk/ena/browser/view/PRJEB45055

unsure if I should report these to identifiers.org or here.. but this issue track seems more active!

bgyori:
Fixes #106.

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -10793,7 +10793,8 @@
     },
     "synonyms": [
       "ena"
-    ]
+    ],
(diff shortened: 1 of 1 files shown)


cthoyt:
I'm so happy about this! being able to override broken stuff in other resources is one of the main reasons for the bioregistry :)

PR #130, opened 2021-07-29
add nmdc to bioregistry

dwinston:
To resolve e.g. `https://bioregistry.io/nmdc:y3ax-8bq3-60` as <https://drs.microbiomedata.org/objects/y3ax-8bq3-60>.

cc @cmungall

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -25408,7 +25408,7 @@
     "name": "National Microbiome Data Collaborative",
     "description": "An initiative to empower the research community to harness microbiome data exploration and discovery through a collaborative integrative data science ecosystem.",
     "homepage": "https://microbiomedata.org/",
-    "example": "jh4z-z81d-76",
+    "example": "y3ax-8bq3-60",
     "pattern": "^[\\w\\-.]{3,}$",
     "url": "https://drs.microbiomedata.org/objects/$1",
     "wikidata": {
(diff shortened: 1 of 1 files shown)


cthoyt:
@dwinston it would be great if you used the template for requesting a new prefix since this will now automatically create a pull request

1. Go to https://github.com/bioregistry/bioregistry/issues/new/choose
2. Click this button Screen Shot 2021-07-29 at 17 14 37
3. It will bring you to the new template page

Just out of curiosity, did you not see this page with all of the links to the templates? Or did you just opt for the blank issue?

cthoyt:
Also I'm not sure where my first comment went... I asked if there's a URL format string that resolves to a human-readable page. If not, then I'll merge

dwinston:
@cthoyt sorry, I didn't see the issue template. I was trying to follow <https://github.com/bioregistry/bioregistry/#-contributing>, and it seemed like the "make edits directly" part meant I should make a pull request directly rather than raise a github issue.

There is not currently a URL format string that resolves to a human-readable page. I can make it so that the given URL format string results in a human-readable page by default, but this will take some time. I hope that it can be merged as is for now. Perhaps I could create another pull request in the future just to change the URL format string if appropriate?

cthoyt:
Yes this is fine! Are you the contact person for this site? There's also a contact field where could put your email if you are.

cthoyt:
Thanks! I think it's up to you decide which will be supported longer / more useful. Technically there's nothing in the schema right now for contact labels, but I think that it might be a good idea to add one. Thanks so much for the contribution! I'll link you when the new version goes live.

PR #138, opened 2021-08-02
Enforce examples for all non-deprecated entries

cthoyt:
This PR does the following:

1. Updates the test for examples to even include entries that don't have a pattern.
2. Add new field in resources for "proprietary" status and assigns a few like EUROFIR and NextMove's NameRXN. There aren't really any metadata resources available for proprietary identifiers, but it's good to have them in the Bioregistry so if people encounter them, they will know what's going on.
3. Allow proprietary resources to not need examples
4. Update new prefix form to enforce it has an example


cthoyt:
16 left!

PR #140, opened 2021-08-02
Add `bioregistry.parse_iri()` functionality

cthoyt:
While being able to turn IRIs into CURIEs wasn't part of the original goals of the bioregistry, it has most of the metadata that would be necessary to implement such a function.

### DIFF
--- a/src/bioregistry/parse_iri.py
+++ b/src/bioregistry/parse_iri.py
@@ -61,7 +61,7 @@ def parse_iri(iri: str) -> Union[Tuple[str, str], Tuple[None, None]]:
     .. todo:: IRI with weird embedding, like ones that end in .html
     """
     if iri.startswith(BIOREGISTRY_PREFIX):
-        curie = iri[len(BIOREGISTRY_PREFIX): ]
+        curie = iri[len(BIOREGISTRY_PREFIX) :]
         return parse_curie(curie)
     if iri.startswith(OLS_URL_PREFIX):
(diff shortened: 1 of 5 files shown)


src/bioregistry/parse_iri.py
```diff
@@ -0,0 +1,84 @@
...
+    .. todo:: IRI with weird embedding, like ones that end in .html
+    """
+    if iri.startswith('https://www.ebi.ac.uk/ols/ontologies/'):
+        return _parse_purl(iri.rsplit('=', 1)[1])
+    if iri.startswith('http://purl.obolibrary.org/obo/TTO_1058367'):
+        return _parse_purl(iri)
+    if iri.startswith('https://identifiers.org/'):
+        return parse_curie(iri[len('https://identifiers.org/'):])
+    if iri.startswith('http://identifiers.org/'):
+        return parse_curie(iri[len('http://identifiers.org/'):])
```
bgyori, line 53: Should we add handling for old-style identifiers.org URLs? For instance http://identifiers.org/uniprot/P15056? These are still widely used.

cthoyt: I thought about this. Probably not a bad idea. I will update the parse curie function to allow for different delimiters, since obo purls also use underscores instead

cthoyt: Implemented in 0f180ed

PR #159, opened 2021-09-20
Enable overriding the prefix map used when parsing and serializing IRIs

cthoyt:
Based on a request from @matentzn to add override context, here's an implementation that accomplishes the following:

```python
from bioregistry import get_iri

# A custom context is always highest priority
context = {"chebi": "https://example.org/chebi/"}
assert get_iri("chebi:24867", context=context) == 'https://example.org/chebi/24867'
assert get_iri("fbbt:1234", context=context, priority=priority) == 'http://purl.obolibrary.org/obo/FBbt_1234'

# A custom context can be supplied in combination with a priority list:
context = {"chebi": "https://example.org/chebi/"}
priority = ["obofoundry", "custom", "default", "bioregistry"]
assert get_iri("chebi:24867", context=context, priority=priority) == 'http://purl.obolibrary.org/obo/CHEBI_24867'
assert get_iri("fbbt:1234", context=context, priority=priority) == 'http://purl.obolibrary.org/obo/FBbt_1234'
```


matentzn:
How can I refer to the custom context in a priority list, say I want first my custom, then obofoundry, then identifiers.org or some such?

cthoyt:
Interesting... this is gonna get tricky but I think it can be done

README.md
```diff
@@ -258,6 +258,16 @@ assert bioregistry.get_iri("chebi:24867", priority=priority) == 'http://purl.obo
...
 ```
 
+Even deeper, you can override any of the bioregistry context directly using
+the `context` keyword:
+
+```python
+from bioregistry import get_iri
+
+context = {"chebi": "https://example.org/chebi/"}
+assert get_iri("chebi:24867", context=context) == 'https://example.org/chebi/24867'
```
cthoyt, line 268: update with nico use case

PR #163, opened 2021-09-23
Add code of conduct

### LINKED ISSUE #161: Adopt contributor covenant
This is a pretty good boilerplate code of conduct that a lot of projects use: https://www.contributor-covenant.org/

For contact I guess I would use my personal email? Or maybe then it's better to make a project-specific email with gmail?

@bgyori thoughts?

cthoyt:
Closes #161 

Questions:

- is this overkill?
- should we make a different email address besides my personal one?

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,133 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, caste, color, religion, or sexual identity
+and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
+* Being respectful of differing opinions, viewpoints, and experiences
+* Giving and gracefully accepting constructive feedback
+* Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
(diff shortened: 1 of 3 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,133 @@
...
+an individual is officially representing the community in public spaces.
+Examples of representing our community include using an official e-mail address,
+posting via an official social media account, or acting as an appointed
+representative at an online or offline event.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported to the community leaders responsible for enforcement at
+cthoyt@gmail.com.
```
cthoyt, line 63: Here's where my personal email has been used to replace the placeholder

cthoyt:
@matentzn have you ever done a code of conduct on any of your projects? Do you think this is overkill? I think the boilerplate from the contributor convenant is a good place to start at least

matentzn:
We have only done the one for OBO:
http://obofoundry.org/docs/COC.html

I think it is overkill for a project like bioregistry at this stage, but I like a reusable coc for _all_ our projects.

cthoyt:
> We have only done the one for OBO:
> http://obofoundry.org/docs/COC.html
> 
> I think it is overkill for a project like bioregistry at this stage, but I like a reusable coc for _all_ our projects.

One reason it might be good to do now is that it's a small project and we haven't established any _real_ governance yet, and it would be easier to make something simple. Later, people doing something like this could require much more effort to get everyone on board.

I'm not sure if you're suggesting that I consider deferring to the general OBO code of conduct as an alternative or just saying that it's good that OBO has it because it's a larger community. If it's the former, I don't think I'd agree - I think the Bioregistry will live its best life neutral towards OBO or any other communities it might serve.

Side note: who's on the OBO COC committee? It's not listed on any of the other pages that have committee memberships.

matentzn:
I dont know who is on the committee, but its somewhere in some document. :D I bet given all the governance work happening this will be formalised and published soon.

I am not saying defer/reuse the OBO one, its just the only one I ever worked on - an example :)

PR #164, opened 2021-09-23
Add data integrity tests for unique IRIs

cthoyt:
This PR adds tests to ensure that when we generate a prefix map using the Bioregistry that all of the IRIs are unique. There are two scenarios we can use to mitigate this I can think of offhand:

1. When entries are in parent/child relationship, like `kegg` with `kegg.compound`, `kegg.drug`, etc. we can enforce that the `part_of` annotation is used for all of the children. If only one prefix in a group of prefixes with the same IRI doesn't have a `part_of` relationship, we'll assume it's the parent and default to using it.
2. When entries are annotated with a `provides`, then they will automatically not be considered in the prefix map, such as `ctd.gene`, which provides for `ncbigene`

When neither of these work, this PR has also introduced the `has_canonical` relationship between prefixes that maps one prefix to another.

There are some entries that seem to be duplicates of each other like `glycomedb` and `glytoucan`. In this case, `glytoucan` is the actual name of the database so `glycomedb` gets the `"has_canonical": "glytoucan"`, denoting it is higher priority (see https://github.com/identifiers-org/identifiers-org.github.io/issues/167).  I wasn't confident enough to merge them, but the `wb` and `wormbase` entries are now indeed merged.

The tests enabled exhaustive curation of scenarios 1, 2, and the `pmapto` is used for the rest.

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -4362,6 +4362,7 @@
       "prefix": "cattleqtldb",
       "provider_url": "http://www.animalgenome.org/cgi-bin/QTLdb/BT/qdetails?QTL_ID=$1"
     },
+    "part_of": "qtldb",
     "prefixcommons": {
       "formatter": "http://identifiers.org/cattleqtldb/$1",
       "is_identifiers": true,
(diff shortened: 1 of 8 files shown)


cthoyt:
@matentzn FYI this is the big step necessary in improving the curation in the Bioregistry to support generating prefix maps that are guaranteed to be bidirectionally unique :)

PR #169, opened 2021-09-28
Add preferred prefix field

### LINKED ISSUE #166: Add `preferred_prefix` field to support broader community
While the Bioregistry already supports the usage of different "profiles" depending on if you're in ontology world (e.g., you want to use OBO PURLs), if you're in the systems biology modelling world (e.g., you want to use Identifiers.org IRIs), etc.

However, the OBO Foundry had the good idea to store both a canonical prefix and a stylized prefix (i.e., the `preferredPrefix` field). It would be nice to add this to the bioregistry as well to allow for writing them in a stylized way for certain downstream uses, especially to capture prefixes like `ncbigene` which often is written as `NCBIGene`, but does not itself appear in the OBO Foundry registry.

cthoyt:
Closes #166

- [x] Implement preferred prefix lookup in Resource class
- [x] Add package-level function
- [x] Amazing docstrings
- [x] Get help doing a bit more curation
- [ ] Write curation guidelines on how to pick these

### DIFF
--- a/src/bioregistry/__init__.py
+++ b/src/bioregistry/__init__.py
@@ -40,6 +40,7 @@ from .resolve import (  # noqa:F401
     get_owl_download,
     get_pattern,
     get_pattern_re,
+    get_preferred_prefix,
     get_prefixcommons_format,
     get_provides_for,
     get_registry_map,
--- a/src/bioregistry/app/templates/resource.html
+++ b/src/bioregistry/app/templates/resource.html
@@ -66,6 +66,11 @@
             <dl>
                 <dt>Prefix</dt>
                 <dd><code>{{ prefix }}</code></dd>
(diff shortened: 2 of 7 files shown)


cthoyt:
it's time to look back into DPO... in OBO foundry it's denoted as having a preferred prefix as FBcv. I mentioned this in https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1559. The comments there said it would be an issue, but I think this is our opportunity to right this wrong

cthoyt:
@cmungall please check this out when you get a chance, and let me know if you've got some guidelines in mind on when people should override the normalized prefixes with a `preferred_prefix`

cmungall (approved):
awesome!

PR #179, opened 2021-09-28
Added a few more Scholia mappings

egonw:


src/bioregistry/data/bioregistry.json
```diff
@@ -40707,4 +40713,4 @@
       "prefix": "ZP"
     }
   }
-}
\ No newline at end of file
+}
```
cthoyt, line 40716: Normally I would agree but this file is managed by a python script so it gets dumped each time with no newline at the end

cthoyt: Fixed in 324dc2c

egonw: I don't even know how to solve this with my editor (`kate`).

egonw: (but I'll make care to use `git add -p` next time)

cthoyt:
@egonw idk why I didn't go with your suggestion before, but it's finally fixed in #2067

PR #194, opened 2021-09-30
Improve getting prefix maps

cthoyt:
The goal of this PR is to make it possible to generate prefix maps that use preferred prefixes (e.g., the nice stylized ones like `FBbt` instead of `fbbt`) instead of the canonical bioregistry prefixes, when available either through OBO Foundry annotations or explicit bioregistry annotations. Specifically it does the following:

1. Put functionality inside manager
2. Enable usage of preferred prefixes
3. Update OBO context to use preferred prefixes
4. Rename some functions to make the package more ergonomic


docs/_data/contexts/obo.context.jsonld
```diff
@@ -47,6 +47,7 @@
         "DIDEO": "http://purl.obolibrary.org/obo/DIDEO_",
         "DINTO": "http://purl.obolibrary.org/obo/DINTO_",
         "DOID": "http://purl.obolibrary.org/obo/DOID_",
+        "DPO": "http://purl.obolibrary.org/obo/FBcv_",
```
cthoyt, line 50: This one needs a better IRI if anyone has an idea

docs/_data/contexts/obo.context.jsonld
```diff
@@ -986,7 +989,6 @@
         "vmhreaction": "https://www.vmh.life/#reaction/",
         "void": "http://rdfs.org/ns/void#",
         "vz": "https://viralzone.expasy.org/",
-        "wb": "https://www.wormbase.org/get?name=",
         "wb.rnai": "https://www.wormbase.org/species/c_elegans/rnai/",
```
cthoyt, line 992: should this one be changed to match `wormbase`?

PR #195, opened 2021-09-30
Scholia can resolve Wikidata Q/P identifiers too

egonw:
@cthoyt, not sure if I did this right. The formatter URL is different, and does not use a prefix in the URL. I tried to learn from the other content, and saw `n2t` under mappings and as separate submap too.

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -42450,8 +42450,7 @@
       "cellosaurus": "Wikidata",
       "miriam": "wikidata",
       "n2t": "wikidata",
-      "prefixcommons": "WD_Entity",
-      "scholia": "scholia"
+      "prefixcommons": "WD_Entity"
     },
     "miriam": {
       "deprecated": false,
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -42513,7 +42513,16 @@
       "is_identifiers": false,
       "is_obo": false,
       "prefix": "WD_Prop"
-    }
+    },
+    "providers": [
+      {
+        "code": "scholia",
+        "description": "Scholia can generally resolve many Wikidata entries in the biomedical and bibliometric domains",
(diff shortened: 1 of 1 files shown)


cthoyt:
@egonw there's actually a prefix request issue template, would you please try filling that one out? It will make sure you get all of the metadata! plus it will automate making a PR for you :)

cthoyt:
plus in the end we will annotate that scholia "provides for" wikidata, and isn't itself a nomenclature authority. Otherwise it might be better to wait on changing the data model to include extra providers

egonw:
> @egonw there's actually a prefix request issue template, would you please try filling that one out? It will make sure you get all of the metadata! plus it will automate making a PR for you :)

But I'm not requesting a prefix...

cthoyt:
Ohh I see what's going on here! This seems to make sense, but let me think about how to incorporate this into the rest of the bioregistry architecture

cthoyt:
@egonw I thought about this a bit more and I'm not sure if it quite fits with the current data model. Check out what happens for the providers:

Screen Shot 2021-10-01 at 12 01 00

This is happening because I added scholia as a registry - meaning that given a prefix/identifier pair, it's possible to construct a URL. The thing you added would be a provider, which knows how to construct a URL given just an identifier, but since this is the only example from scholia, I didn't set it up to handle this.

I just made a new issue https://github.com/biopragmatics/bioregistry/issues/196 to establish a data model for additional providers - this would be the place to add more annotations. I will make a prototype then help update this PR

cthoyt:
@egonw I've now spent a few hours on preparing the Bioregistry for this kind of curation, and it's ready for prime-time! I updated your PR a bit, and now it's possible to add extra providers with some associated metadata. Maybe it would be good to make schema.org classes for these objects so it's not the responsibility of the Bioregistry to dictate what has to be included to talk about a "provider"?

Anyway, here's how the Wikidata page looks now (it's a bit small but I squeezed everything in there)

Screen Shot 2021-10-01 at 15 56 30

PR #202, opened 2021-10-03
Add INDRA resolution

cthoyt:
This PR adds the programmatically generated INDRA database URLs for mappable prefixes. Currently it's got manually mapped prefixes, but there's a mapping living somewhere that we can consume to keep this up to date (hopefully without having to install/import INDRA).

see `from indra.databases.identifiers import identifiers_mappings`

Screen Shot 2021-10-03 at 18 41 39

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -17253,6 +17253,7 @@
       "bioportal": "HGNC",
       "cellosaurus": "HGNC",
       "go": "HGNC",
+      "indra": "HGNC",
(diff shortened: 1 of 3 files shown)


bgyori:
I'm actually not sure this is something we'd want. Shouldn't resolutions primarily be for landing pages representing a given entity in a primary resource? (As opposed to some arbitrary database providing data linked to a given entity)

cthoyt:
After some more consideration I agree with Ben. In the mean time, we've implemented some features for adding arbitrary providers which might be more appropriate for including links to the INDRA database

PR #210, opened 2021-10-10
Add assorted antibody vendors

cthoyt:
This PR adds some antibody vendors to support generating structured cross-references from the Antibody Registry (see https://github.com/pyobo/pyobo/pull/121)


bgyori:
This may or may not be useful but CiteAb is a pretty useful antibody aggregator. The IDs have an ad-hoc/complicated format though, e.g., `125835-9165-c-jun-60a8-rabbit-mab` -> https://www.citeab.com/antibodies/125835-9165-c-jun-60a8-rabbit-mab.

cthoyt:
Thanks! I was working my way down from the most frequent ones showing up in antibody registry but I'll be sure to get that one since you mentioned it

cthoyt:
@bgyori I looked back into this and the resolution of CiteAb is really nonsense. it looks like there are accession numbers, like `9165` for your example that had address https://www.citeab.com/antibodies/125835-9165-c-jun-60a8-rabbit-mab but I can't figure out if there's a more meaningful redirect service

Issue #232, opened 2021-11-01
Better documentation of the registry data files


dhimmel:
From the [readme](https://github.com/biopragmatics/bioregistry/blob/e75275160e6b6ffd619e83579bbc629e3d0309f6/README.md#-download):

> The underlying data of the Bioregistry can be downloaded directly
from [here](https://github.com/biopragmatics/bioregistry/blob/main/src/bioregistry/data/bioregistry.json).
Several exports to YAML, TSV, and RDF can be downloaded via https://bioregistry.io/download.

Looking through the repo, I see the following two locations for data:
 
- `src/bioregistry/data` presumably for input data
- `docs/_data` presumably for output data (generated by CI)

But I don't see any docs for these files, such as a README file.

In https://github.com/manubot/manubot/issues/305, @cthoyt pointed me to `docs/_data/registry.tsv` which contains what I want: the actual bioregistry metadata fields after merging all the input registries (i.e. the resolved/consensus view). But `docs/_data/registry.yaml` seems not to contain all the output fields (despite having the same file stem name).

This issue has two parts:

1. better documentation of the data, especially the output files
2. based on 1, a request to create a JSON/YAML format of the consensus view available in `docs/_data/registry.tsv`

bgyori:
I agree, we should work on these, thanks!

cthoyt:
[Consensus JSON](https://github.com/biopragmatics/bioregistry/blob/main/exports/registry/registry.json) and [consensus YAML](https://github.com/biopragmatics/bioregistry/blob/main/exports/registry/registry.yml) files are available as of #235 along with some nice documentation updates. #236 will update the documentation of the internal data then we can see if we've addressed all of @dhimmel's concerns

cthoyt:
Sorry for the double message but I wanted to get this all cleaned up and merged. If there's any feedback or follow-up that you think we should do after #235 and #236, please let us know :)

dhimmel:
Fantastic! Changes look great and much clearer now what data is where. Looking forward to using `exports/registry/registry.json`.

PR #235, opened 2021-11-02
Generate consensus view of registry and reorganize all exports

cthoyt:
References #232 

This PR does the following:

1. Moves the exports out of the `docs/_data/` folder and into the `exports/` folder
2. Adds one level of folder hierarchy to group exports of the same resource
3. Adds README files describing what each file is and where it comes from
4. Adds consensus view in both JSON and YAML for the registry (this isn't necessary for the other resources for the metaregistry and collections)
5. Update the global README and web app's downloads page to reflect new file paths


cthoyt:
@matentzn @cmungall heads up, this PR resulted in moving the OBO JSON-LD context file to a new URL. Maybe we should get PURLs for these resources?

PR #239, opened 2021-11-02
Remove more requirements from standard install

cthoyt:


bgyori:
I don't want to make this too complicated, just curious, aren't some of the remaining dependencies like rdflib just needed for custom exports and could be made optional?

cthoyt:
@bgyori yes, sort of. The only issue is the bits with RDFlib are currently scattered all over and when I was thinking about how to make them more optional, it would cause a lot of code to get *really* hacky. Will reconsider this before merging.

cthoyt:
So i looked into it and actually was able to do it without making too much of a mess, but I'll make a separate PR since it does modify quite a few files

PR #243, opened 2021-11-03
Fix pattern checking and fix panther.pthcmp pattern

### LINKED ISSUE #242: Update regular expression pattern for panther.pthcmp to match example identifier
### Prefix

panther.pthcmp

### New Regular Expression Pattern

^(G|P|U|C|S)\d{5}$

### Explanation

The pattern for `panther.pthcmp` (`^G|P|U|C|S\\d{5}$`) doesn't match the example `P00266`:

https://github.com/biopragmatics/bioregistry/blob/e9b76c42dbc5cd0e5552093427b2f0a7bbe3b046/exports/registry/registry.json#L14169-L14182

Side note: I am surprised that patterns that don't match the example are allowed. Would think CI should flag this and suppress the pattern/example fields until this is resolved?

### Contributor ORCID

0000-0002-3012-7446

cthoyt:
Closes #242

### DIFF
--- a/src/bioregistry/schema/struct.py
+++ b/src/bioregistry/schema/struct.py
@@ -949,7 +949,10 @@ class Resource(BaseModel):
         pattern = self.get_pattern_re()
         if pattern is None:
             return None
-        return bool(pattern.match(identifier))
+        match = pattern.match(identifier)
+        if not match:
+            return False
+        return match.group(0) == identifier
 
     def is_known_identifier(self, identifier: str) -> Optional[bool]:
         """Check that a local unique identifier can be normalized and also matches a prefix's pattern."""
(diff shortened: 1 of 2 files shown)


src/bioregistry/schema/struct.py
```diff
@@ -949,7 +949,7 @@ def is_canonical_identifier(self, identifier: str) -> Optional[bool]:
         pattern = self.get_pattern_re()
         if pattern is None:
             return None
-        return bool(pattern.match(identifier))
+        return pattern.fullmatch(identifier) is not None
```
dhimmel, line 952: Pretty interesting case of how match can easily go wrong when there are conditionals in the regex, since the `$` doesn't get included in every option.

fullmatch is a nice solution!

dhimmel (commented):
Approve

cthoyt:
@dhimmel just minted release v0.4.3 with this update on PyPI

PR #248, opened 2021-11-08
Add resolution URL override for Uberon

bgyori:
The current resolution URL for Uberon (imported via Miriam, redirects to a URL of the form https://bioportal.bioontology.org/ontologies/UBERON?p=classes&conceptid=UBERON%3A2005080) is not working. This PR adds an override to link to e.g., https://bioportal.bioontology.org/ontologies/UBERON/?p=classes&conceptid=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FUBERON_2005080 which resolves correctly.

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -41456,6 +41456,7 @@
       "Uberon",
       "Uber-anatomy ontology"
     ],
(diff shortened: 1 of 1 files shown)


cthoyt:
@bgyori I thought about this for a bit, and maybe that it's the underlying OBO PURL system that needed a fix. But further investigation showed that it's pulling this URI formatter from Identifiers.org, and an alternative technical solution would be more appropriate where BioPortal links aren't propagated from Identifiers.org if an OBO PURL is available

bgyori:
Yes, working, thanks!

cthoyt:
Great, then I'm going to close this one

Issue #262, opened 2021-12-09
What prefix to denote a database as an entity?


cmungall:
Carried on from:
https://github.com/mapping-commons/sssom/issues/126

A common issue is being able to refer to a particular database or data source as an authority, as if it were an CURIE's entity in and of itself

We can use bioregistry: as a prefix to denote sources, e.g. bioregistry:uberon, bioregistry:pubmed

Alternatively, we could use idot, e.g. identifiers.namespace:uberon, identifiers.namespace:pubmed

Bioregistry will happily redirect either

 * https://bioregistry.io/bioregistry:pubmed 
 * https://bioregistry.io/identifiers.namespace:pubmed _redirects to_ https://registry.identifiers.org/registry/pubmed

However, idot will not resolve bioregistry:pubmed

One position is that `bioregistry:pubmed` and `identifiers.namespace:pubmed` denote *registry entries for databases* and not databases themselves. If one wants to refer to a database then one should use a dedicated database of databases, including

 - Wikidata
 - ROR? (organization is often but not always synonymous with database)
 - Infores (not yet registered with bioregistry)
 - Fairsharing, e.g https://fairsharing.org/biodbcore-000456 for pubmed -- does not yet work with bioregistry

Another position is that adopting `bioregistry:DATABASE` as a convention works pragmatically and we know we have good coverage of databases and it's easy to add new ones.

I think the main argument against using bioregistry as the prefix is that it is inherently limited to ID-minting sources. For example, we may want to register "monarch" as the curation source of some annotations, but I don't think we would want to register monarch in bioregistry, at least until we start minting IDs (which we might in the future - e.g ids for annotations)

OK to close as out of scope

matentzn:
Before closing, I would at least like to know @cthoyt position on this. I don't mind resorting to Wikidata, but, the appeal of using a prefixed ID with a _readable_ source name like br:UBERON has a lot of appeal.. I don't like the fact that for these sort of things, where the identifier space is extremely finite, we have to resort to opague numeric ids..

cthoyt:
Bioregistry identifiers correspond to nomenclatures, not databases. 
While there's a big overlap in those things, they're not exactly the same. See https://cthoyt.com/2021/10/07/biopragmatics-glossary.html for a few more thoughts on that.

[...]

PR #265, opened 2021-12-09
Automatically upgrade provider prefixes

cthoyt:
Motivated by #262


cthoyt:
This needs a bit more thinking, so discussion will continue in #758.

PR #288, opened 2022-01-15
Update DOI entry

### LINKED ISSUE #287: doi and linked data URLs - dx vs non-dx
http://bioregistry.io/registry/doi

will map a CURIE `doi:10.1038/nbt1156` to 
http://doi.org/10.1038/nbt1156 

which is perfectly fine for humans (although it should be https, that is not my concern here)

what URI should be used in a linked data context?

Let's ask doi:

```turtle
✗  curl -s -LH "Accept: text/n-triples" http://doi.org/10.1038/nbt1156  | grep sameAs
<http://dx.doi.org/10.1038/nbt1156> <http://www.w3.org/2002/07/owl#sameAs> <info:doi/10.1038/nbt1156> .
<http://dx.doi.org/10.1038/nbt1156> <http://www.w3.org/2002/07/owl#sameAs> <doi:10.1038/nbt1156> .
<http://id.crossref.org/issn/1087-0156> <http://www.w3.org/2002/07/owl#sameAs> "urn:issn:1087-0156" .
<http://id.crossref.org/issn/1087-0156> <http://www.w3.org/2002/07/owl#sameAs> "urn:issn:1546-1696" .
```

They are explicitly declaring a clique of 3 equivalent URIs. However, none of them are `//doi.org`, two an URNs, and the only URL is `http://dx.doi.org/` based

This yields nothing:

```bash
curl -s -LH "Accept: text/n-triples" http://doi.org/10.1038/nbt1156  | grep //doi.org
```

I believe dx is widely regarded as supported but not preferred. However, this is in contrast to the triples provided.

I found this factsheet
https://www.doi.org/factsheets/DOI_PURL.html

But it seems to be an out-of-context argument against using the purl.org server?

Ideally bioregistry could be a clearinghouse for what the best linked data identifier to use is for a resource like DOIs. It seems we have centered on the non-dx form for most OBO ontologies. Which is in line with conventional wisdom. But it's troubling that this is not the same as the semantic web IDs produced by the crossref system itself.

Maybe there is no action here, and it is what it is

cthoyt:
Closes #287 CC @cmungall 

This PR does several things:

1. Uses the HTTP / dx.doi.org URI format string for the [`doi`](https://bioregistry.io/registry/doi) prefix
2. Adds three alternate providers into the DOI entry corresponding the the HTTPS / dx.doi.org URI format string and both variants of the doi.org URI format string.

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -11931,7 +11931,31 @@
       "is_obo": false,
       "prefix": "DOI",
       "uri_format": "http://dx.doi.org/$1"
-    }
+    },
+    "providers": [
+      {
+        "code": "dx_doi_https",
+        "description": "An alternate provider from the DOI website using HTTPS",
+        "homepage": "https://www.doi.org",
+        "name": "Digital Object Identifier",
+        "uri_format": "https://dx.doi.org/$1"
+      },
+      {
+        "code": "doi_https",
+        "description": "An alternate provider from the DOI website",
+        "homepage": "https://www.doi.org",
+        "name": "Digital Object Identifier",
+        "uri_format": "https://doi.org/$1"
+      },
+      {
+        "code": "doi_http",
+        "description": "An alternate provider from the DOI website",
+        "homepage": "https://www.doi.org",
+        "name": "Digital Object Identifier",
+        "uri_format": "http://doi.org/$1"
+      }
+    ],
+    "uri_format": "http://dx.doi.org/$1"
   },
   "doid": {
     "bioportal": {


src/bioregistry/data/bioregistry.json
```diff
@@ -11931,7 +11931,31 @@
...
+      },
+      {
+        "code": "doi_http",
+        "description": "An alternate provider from the DOI website",
+        "homepage": "https://www.doi.org",
+        "name": "Digital Object Identifier",
+        "uri_format": "http://doi.org/$1"
+      }
+    ],
+    "uri_format": "http://dx.doi.org/$1"
```
dhimmel, line 11958: I like the previous uri format for DOIs:

```suggestion
    "uri_format": "https://doi.org/$1"
```

Since that is supposedly "preferred" at least for humans, but I guess not for linked data as per https://github.com/biopragmatics/bioregistry/issues/287.

cmungall:
I guess as sort of a linked data person and the originator of the ticket I
would say just go with the canonical one Daniel recommends it’s what we use
in all our owl anyway… my point with the original ticket was to highlight
the big difference between what doi were saying with one half of their
stack and the other and maybe even start a dialog with them to get their
triples changed

src/bioregistry/data/bioregistry.json
dhimmel: > In https://github.com/biopragmatics/bioregistry/pull/291, I'm trying to make the alternate resolvers more accessible for generating URIs

Sounds like a good idea, although the challenge seems to be that there are two uses for the URI:

1. as a web address that users can go to. These should do things like use HTTPS when possible and be human readable if possible.
2. as a URI identifier for linked data where following the standard is most important even if that results in a legacy (or non-operational) URL.

It's impractical for the user to know which provider achieves which objective for every prefix. So what would be ideal is if the user could do something like: `bioregistry.get_iri("prefix:id", usage="linked_data")` or `bioregistry.get_iri("prefix:id", usage="webpage")`. This would require major work, not saying it's worth it.

PR #293, opened 2022-01-25
Fix bioRxiv URI formatter, add regular expression pattern, and add extra examples

dhimmel:
refs https://github.com/manubot/manubot/issues/282

The example is `2021.07.23.453588`. The following URL resolves this identifier: https://www.biorxiv.org/content/10.1101/2021.07.23.453588. The current `uri_format` results in a broken URL.

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -4133,6 +4133,7 @@
   "biorxiv": {
     "description": "The bioRxiv is a preprint server for biology",
     "example": "2021.07.23.453588",
+    "pattern": "^\\d{4}\\.\\d{2}\\.\\d{2}\\.)?\\d{6}(v\\d{1,3})?$",
     "go": {
       "homepage": "https://www.biorxiv.org/",
       "name": "bioRxiv",
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -4133,7 +4133,7 @@
   "biorxiv": {
     "description": "The bioRxiv is a preprint server for biology",
     "example": "2021.07.23.453588",
-    "pattern": "^\\d{4}\\.\\d{2}\\.\\d{2}\\.)?\\d{6}(v\\d{1,3})?$",
+    "pattern": "^(\\d{4}\\.\\d{2}\\.\\d{2}\\.)?\\d{6}(v\\d{1,3})?$",
     "go": {
       "homepage": "https://www.biorxiv.org/",
       "name": "bioRxiv",
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -4134,6 +4134,11 @@
     "description": "The bioRxiv is a preprint server for biology",
     "example": "2021.07.23.453588",
     "pattern": "^(\\d{4}\\.\\d{2}\\.\\d{2}\\.)?\\d{6,8}(v\\d{1,3})?$",
+    "example_extras": [
+      "000091",
+      "000091v1",
+      "2022.01.24.477526v1"
+    ],
     "go": {
       "homepage": "https://www.biorxiv.org/",
       "name": "bioRxiv",
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -4140,7 +4140,7 @@
       "synonyms": [
         "BIORXIV"
       ],
-      "uri_format": "https://www.biorxiv.org/content/$1"
+      "uri_format": "https://www.biorxiv.org/content/10.1101/$1"
     },
     "homepage": "https://biorxiv.org",
     "mappings": {
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -4133,7 +4133,7 @@
   "biorxiv": {
     "description": "The bioRxiv is a preprint server for biology",
     "example": "2021.07.23.453588",
-    "pattern": "^(\\d{4}\\.\\d{2}\\.\\d{2}\\.)?\\d{6}(v\\d{1,3})?$",
+    "pattern": "^(\\d{4}\\.\\d{2}\\.\\d{2}\\.)?\\d{6,8}(v\\d{1,3})?$",
     "go": {
       "homepage": "https://www.biorxiv.org/",
       "name": "bioRxiv",
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -4133,7 +4133,6 @@
   "biorxiv": {
     "description": "The bioRxiv is a preprint server for biology",
     "example": "2021.07.23.453588",
-    "pattern": "^(\\d{4}\\.\\d{2}\\.\\d{2}\\.)?\\d{6,8}(v\\d{1,3})?$",
     "example_extras": [
       "000091",
       "000091v1",
@@ -4153,7 +4152,8 @@
       "go": "bioRxiv",
       "scholia": "biorxiv"
     },
-    "name": "bioRxiv"
+    "name": "bioRxiv",
+    "pattern": "^(\\d{4}\\.\\d{2}\\.\\d{2}\\.)?\\d{6,8}(v\\d{1,3})?$"
   },
   "biosample": {
     "cellosaurus": {


src/bioregistry/data/bioregistry.json
```diff
@@ -4140,7 +4140,7 @@
       "synonyms": [
         "BIORXIV"
```
dhimmel, line 4141: Curious why there is an ALL CAPS synonym? Shouldn't synonyms be distinct in a case-insensitive way... i.e. a `BIORXIV` prefix is already supported without this.

cthoyt: I've curated the synonyms in a case-sensitive way - so this reflects what I've seen in the wild. Internally, the bioregistry throws away case sensitivity when it does lookup.

cthoyt:
@dhimmel do you think while you're at it, you could annotate a regular expression pattern for bioRxiv entries as well?

dhimmel:
Currently [wikidata](https://www.wikidata.org/wiki/Property:P3951) has the following: `(\d{4}\.\d{2}\.\d{2}\.)?\d{6}`. Seems like we should also add the optional version suffix. Will update PR.

Also noting the [bioRxiv announcement](https://twitter.com/cshperspectives/status/1205109455752454144) of the date prefixed format.

src/bioregistry/data/bioregistry.json
```diff
@@ -4133,14 +4133,15 @@
   "biorxiv": {
     "description": "The bioRxiv is a preprint server for biology",
     "example": "2021.07.23.453588",
+    "pattern": "^\\d{4}\\.\\d{2}\\.\\d{2}\\.)?\\d{6}(v\\d{1,3})?$",
```
dhimmel, line 4136: Regarding the regex pattern, I allowed version suffixes up to `v999`. Assuming a preprint will never have 1000 versions. Possibly 100 would be safe, but not in a future where preprints are updated automatically on a scheduled basis.

Am worried that the main 6 digit number will eventually be exceeded. A recent ID is `2022.01.24.477526v1`, suggesting that there have been `477526` preprints and eventually this might exceed 1 million.

Should we keep this at 6 for now and fix in the future, or proactively allow up to 7 or 8 digits?

cthoyt: I think it’s alright to just have an arbitrary \d in this spot

cthoyt: But then again it’s nice to have extra guarantees for funky looking data, so maybe just go one order of magnitude higher than what’s there already

dhimmel: Made 6-8 digits in https://github.com/biopragmatics/bioregistry/pull/293/commits/78dd4e2e3664df7c7a3ecad04174027d5f65a4ac. Seems like if we're to hit 1 million preprints, we could also soon hit 10 million.

By the way, the smallest ID I could find for a biorxiv preprint is `000091`. I do like catching errors like `biorxiv:91` which will fail to resolve.

dhimmel: > By the way, the smallest ID I could find for a biorxiv preprint is 000091.

Looking through the Nov 2013 archives, there is also `000067`

src/bioregistry/data/bioregistry.json
```diff
@@ -4133,14 +4133,15 @@
   "biorxiv": {
     "description": "The bioRxiv is a preprint server for biology",
     "example": "2021.07.23.453588",
```
dhimmel, line 4135: Side thought: it would be nice to be able to provide multiple examples for situations where there are different subtypes of IDs.

cthoyt: I was just thinking about that while I was AFK! I have two ideas on how to do it. I will make a PR to try it out

cthoyt: Done in #296

cthoyt:
@dhimmel I just merged in #296 which added support for multiple examples (both in the schema and the automated testing framework). Please let me know if you're happy with the examples I added, then I can merge this.

dhimmel:
> Please let me know if you're happy with the examples I added, then I can merge this.

Looks good!

cthoyt:
@dhimmel thank you as always for the contribution! I think this proves that having an open database where people can make fixes as they notice they're necessary is a good thing

dhimmel:
This PR was merged in a2c55ac6029d42ac1b4591eba391c4b82d959c67. Looking at the commit that updated the exports following the merge https://github.com/biopragmatics/bioregistry/commit/42b2bc7670be9cbbd97113ba3a536b0a1d8f0b0e, `uri_format` did not update:

```patch
diff --git a/exports/registry/registry.yml b/exports/registry/registry.yml
index 8af4b692..bcdba4a7 100644
--- a/exports/registry/registry.yml
+++ b/exports/registry/registry.yml
@@ -1874,11 +1874,16 @@ bioregistry.schema:
 biorxiv:
   description: The bioRxiv is a preprint server for biology
   example: 2021.07.23.453588
+  example_extras:
+  - 000091
+  - 000091v1
+  - 2022.01.24.477526v1
   homepage: https://biorxiv.org
   mappings:
     go: bioRxiv
     scholia: biorxiv
   name: bioRxiv
+  pattern: ^(\d{4}\.\d{2}\.\d{2}\.)?\d{6,8}(v\d{1,3})?$
   preferred_prefix: biorxiv
   uri_format: https://www.biorxiv.org/content/$1
```

Any idea why not?

src/bioregistry/data/bioregistry.json
```diff
@@ -4133,21 +4133,27 @@
...
+      "uri_format": "https://www.biorxiv.org/content/10.1101/$1"
     },
     "homepage": "https://biorxiv.org",
     "mappings": {
       "go": "bioRxiv",
       "scholia": "biorxiv"
     },
-    "name": "bioRxiv"
+    "name": "bioRxiv",
+    "pattern": "^(\\d{4}\\.\\d{2}\\.\\d{2}\\.)?\\d{6,8}(v\\d{1,3})?$"
```
cthoyt, line 4156: @dhimmel it looks like the pattern was updated here

cthoyt:
Not sure what the issue with the commit hashes were (maybe because I use squash?) but the pattern did indeed make it into the main branch

dhimmel:
> the pattern did indeed make it into the main branch

The merge went fine. The `uri_format` is correct in `src/bioregistry/data/bioregistry.json`. But the new `uri_format` didn't make it to the auto-generated export in `exports/registry/registry.yml`.

dhimmel:
Checking the current output, it looks the `uri_format` still hasn't updated to the version from this PR:

https://github.com/biopragmatics/bioregistry/blob/5ab4e47654a5e8ff8ed6c20ad18e4fb9c243db86/exports/registry/registry.yml#L1925-L1939

@cthoyt can you look into what's going on?

src/bioregistry/data/bioregistry.json
```diff
@@ -4133,21 +4133,27 @@
...
+    ],
     "go": {
       "homepage": "https://www.biorxiv.org/",
       "name": "bioRxiv",
       "prefix": "bioRxiv",
       "synonyms": [
         "BIORXIV"
       ],
-      "uri_format": "https://www.biorxiv.org/content/$1"
+      "uri_format": "https://www.biorxiv.org/content/10.1101/$1"
```
cthoyt, line 4148: Ahhhh I totally missed this, you edited the imported URI format from GO, not the top-level record. Will fix no!

cthoyt:
Fixed the issue in 9d8bf7c. Do you need a re-deploy ASAP or can you wait for the automated one tonight?

src/bioregistry/data/bioregistry.json
cthoyt: Fixed in https://github.com/biopragmatics/bioregistry/commit/9d8bf7cb97f43e5af5204a89358fde5385fc185f

dhimmel: Ah shoot! My bad. Didn't realize this file had the imported registry records.

dhimmel:
> Do you need a re-deploy ASAP or can you wait for the automated one tonight?

Wait for automated.

cthoyt:
Sorry for the long wait on this, I looked into this and somehow the change got lost. I just checked on the live site that it's working properly.

PR #312, opened 2022-02-07
Add Nico's updates

### LINKED ISSUE #306: pav prefix has strange ideosyncractic prefix resolution
Check http://bioregistry.io/registry/pav

When getting a context, I expect

```
pav: http://purl.org/pav/
```

instead, I get

```
pav: https://pav-ontology.github.io/pav/index.html#http://purl.org/pav/
```

http://purl.org/pav/authoredBy, the official IRI, redirects to an RDF dump of PAV (should have used content negotiation).

For the sssom context, i need:

```
pav: http://purl.org/pav/
```

Would that be possible?

### LINKED ISSUE #307: Should "dc:" prefix default to elements or terms?
I am generally using `dce` for elements, and `dcterms` for terms, but its not clear to me how to decide whether the one or the other is preferable for `dc`

- http://purl.org/dc/terms/
- http://purl.org/dc/elements/1.1/

Bioregistry suggests element. DC itself tries to promote terms.. Whats you stance on that @cthoyt?

### LINKED ISSUE #308: Prov namespace should be http, not https
As per official documentation: https://www.w3.org/TR/prov-o/

image

### LINKED ISSUE #309: neuronames prefix should resolve to http, not https
Else it wont resolve..

http://braininfo.rprc.washington.edu/centraldirectory.aspx?ID=407

https://braininfo.rprc.washington.edu/centraldirectory.aspx?ID=407

cthoyt:
Closes #306, closes #307, closes #308, closes #309

@matentzn please let me know if these all look okay - not sure if I hit all points for DC for #307

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -37029,7 +37029,7 @@
       "name": "Charles Tapley Hoyt",
       "orcid": "0000-0003-4423-4370"
     },
-    "uri_format": "https://pav-ontology.github.io/pav/index.html#http://purl.org/pav/$1"
+    "uri_format": "http://purl.org/pav/$1"
   },
   "paxdb.organism": {
     "mappings": {
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -40569,7 +40569,7 @@
       "prefix": "prov",
       "uri_format": "http://www.w3.org/ns/prov#$1"
     },
-    "uri_format": "https://www.w3.org/ns/prov#$1"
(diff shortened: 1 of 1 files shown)


matentzn:
I would suggest adding [this](https://github.com/biopragmatics/bioregistry/issues/307#issuecomment-1031794339), else it looks great!

cthoyt:
@matentzn since we now have a URI format clash, we have to decide which of DC and DCTERMS is the preferred entry (even if we don't merge them). Which do you like better?

matentzn:
`dcterms` is the least ambiguous.. I personally prefer that.

PR #316, opened 2022-02-12
Use preferred DOI uri_format

dhimmel:
refs https://github.com/biopragmatics/bioregistry/issues/287
reverts part of https://github.com/biopragmatics/bioregistry/pull/288

The [DOI resolution factsheet](https://www.doi.org/factsheets/DOIProxy.html) specifies that `https://doi.org/DOI` is the preferred format:

> Users may resolve DOI names that are structured to use the DOI system Proxy Server (https://doi.org (preferred)). The resolution of the DOI name in this case depends on the use of URL syntax: the example DOI name doi:10.10.123/456 would be resolved from the address: "https://doi.org/10.123/456". Any standard browser encountering a DOI name in this form will be able to resolve it. The proxy service (both doi.org and the **earlier but no longer preferred** dx.doi.org) is accessible over IPv6, and supports DNSSEC. The proxy servers respond to HTTPS (**preferred**) as well as HTTP requests.

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -11915,6 +11915,7 @@
       "prefix": "DOI",
       "uri_format": "https://dx.doi.org/$1"
     },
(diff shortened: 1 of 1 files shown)


dhimmel:
@cthoyt review?

cthoyt:
Hi @dhimmel thanks for being patient. I'm going to put a note inside this entry to help people who might wonder why it's set up this way then merge this.

cthoyt:
In a follow-up pull request, I also added markdown support so this gets better shown in the service. That should end up on the main site on the nightly build tonight

PR #318, opened 2022-02-16
Add SHACL Export

### LINKED ISSUE #305: Use SHACL vocabulary in RDF
[SHACL](https://www.w3.org/TR/shacl/) is a controlled vocabulary that has the facility for describing the structure of RDF itself.

The OBO Foundry has demonstrated it's useful to represent a prefix map using SHACL. Their implementation that converts the OBO Foundry's prefix map into SHACL is here https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/util/make-shacl-prefixes.py

It would be useful to create a similar export of the Bioregistry. This could be implemented in this file: https://github.com/biopragmatics/bioregistry/blob/main/src/bioregistry/export/prefix_maps.py next to the code that generates JSON-LD-style prefix maps.

kkaris:
Closes #305.

Remaining issues:
- [x] is storage location OK? https://github.com/biopragmatics/bioregistry/commit/166da5633647f805dee2a71fba139eb031b9d8f7
- [x] ~Is prefix uppercase OK?~ **Fixed** _See [here](https://github.com/kkaris/bioregistry/blob/issue-305/src/bioregistry/export/prefix_maps.py#L55)_
- [x] Add README in `shacl` folder explaining what this is (just copy text from the issue and link to the issue, see the other READMEs for exports to see an example)


cthoyt:
- Don't uppercase
- Storage location looks good!
- Please run this CLI and commit the results in this PR
- We'll need a readme in that folder explaining what the output is

src/bioregistry/export/prefix_maps.py
```diff
@@ -34,6 +35,30 @@ def generate_context_json_ld():
             json.dump(fp=file, indent=4, sort_keys=True, obj=get_collection_jsonld(key))
 
 
+@click.command()
+def generate_shacl_prefixes():
```
cthoyt, line 39: just put this inside the already existing main function, perhaps we need to rename it to be more general though

src/bioregistry/export/prefix_maps.py
```diff
@@ -15,25 +15,56 @@
...
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+[
+    sh:declare
+{entries}
+] ."""
+    # Todo: Should we really uppercase prefixes?
+    path.parent.mkdir(parents=True, exist_ok=True)
+    entries = ",\n".join(
+        f'    [ sh:prefix "{prefix.upper()}" ; sh:namespace "{url}" ]'
+        for prefix, url in prefix_map.items()
```
cthoyt, line 62: rename `url` to `uri_format`

kkaris: Fixed in https://github.com/biopragmatics/bioregistry/pull/318/commits/ef82d08b43d2cbea957a7e09e366e12fc9fc6759

src/bioregistry/export/prefix_maps.py
kkaris: Fixed in https://github.com/biopragmatics/bioregistry/pull/318/commits/6f8b0e0749b553c03c187fc01825ec997c646022

cthoyt (approved):
Looks great, thanks @kkaris :shipit:

Issue #320, opened 2022-02-18
CI fails on forks: config.annotations missing


dhimmel:
CI is [failing on my fork](https://github.com/dhimmel/bioregistry/runs/5241512685?check_suite_focus=true#step:5:21):

```python-traceback
[01:20:[23](https://github.com/dhimmel/bioregistry/runs/5241512685?check_suite_focus=true#step:5:23)] Aligning ols
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/runner/work/bioregistry/bioregistry/src/bioregistry/__main__.py", line 8, in <module>
    main()
  File "/home/runner/work/bioregistry/bioregistry/.tox/update/lib/python3.9/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/home/runner/work/bioregistry/bioregistry/.tox/update/lib/python3.9/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/home/runner/work/bioregistry/bioregistry/.tox/update/lib/python3.9/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/runner/work/bioregistry/bioregistry/.tox/update/lib/python3.9/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/runner/work/bioregistry/bioregistry/.tox/update/lib/python3.9/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/home/runner/work/bioregistry/bioregistry/.tox/update/lib/python3.9/site-packages/click/decorators.py", line [26](https://github.com/dhimmel/bioregistry/runs/5241512685?check_suite_focus=true#step:5:26), in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/runner/work/bioregistry/bioregistry/src/bioregistry/cli.py", line [34](https://github.com/dhimmel/bioregistry/runs/5241512685?check_suite_focus=true#step:5:34), in update
    ctx.invoke(align)
[...]

PR #321, opened 2022-02-19
CI: skip scheduled builds on forks

dhimmel:
refs https://github.com/biopragmatics/bioregistry/issues/320

I was getting a bunch of emails from scheduled workflows failing on my fork. I don't think the scheduled workflows are relevant to forks. Workflow dispatch of these workflows on forks should still be possible.


dhimmel:
CC @cthoyt for review.

cthoyt:
Thanks @dhimmel :)

PR #327, opened 2022-03-02
Add Wikidata database to `4dn.biosource`

CondeArcana333:
good morning bioresigry, we are the bioinformation laboratory( https://sites.google.com/a/ciencias.unam.mx/layla-michan/Home?authuser=0 ) , chairman by the researcher layla michán aguirre, we will be curating and registering the wikidata database.
This is the first addition we make. create the 4dn.biosource database record in wikidata and add it to the bioregistry.json file

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -96,10 +96,9 @@
     "synonyms": [
       "4DN"
     ],
-    "uri_format": ["https://data.4dnucleome.org/biosources/$1"],
+    "uri_format": "https://data.4dnucleome.org/biosources/$1",
     "wikidata": {
       "database": "Q110613621"
-
     }
   },
   "4dn.replicate": {
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -96,7 +96,11 @@
     "synonyms": [
       "4DN"
     ],
-    "uri_format": "https://data.4dnucleome.org/biosources/$1"
+    "uri_format": ["https://data.4dnucleome.org/biosources/$1"],
+    "wikidata": {
+      "database": "Q110613621"
+
+    }
   },
   "4dn.replicate": {
     "contributor": {


cthoyt:
@CondeArcana333 this is a wonderful contribution, thank you!

I didn't realize that P1343 was appropriate for connecting entries in Wikidata. Ideally, we'd mint a property like [identifiers.org prefix (P4793)](https://www.wikidata.org/wiki/Property:P4793) but this is a bit of a painful process. In the mean time, this query can show all entries connected via P1343: https://w.wiki/4uJy

Please let me know if you plan to make more contributions within this PR or if you'd like to get it merged now.

Final note - not sure why you changed the URI format to be a list. This is not the appropriate datatype for this field and the tests will fail. You can either change it back or I'll take care of it :)

CondeArcana333:
good afternoon, I would like to merge it now. I changed the url format just because it marked me an error if I didn't do it and I didn't know how to correct it, if you can take care of that I will be very grateful and in the following contributions I will be more careful with it.

src/bioregistry/data/bioregistry.json
```diff
@@ -96,7 +96,11 @@
     "synonyms": [
       "4DN"
     ],
-    "uri_format": "https://data.4dnucleome.org/biosources/$1"
+    "uri_format": ["https://data.4dnucleome.org/biosources/$1"],
```
cthoyt, line 99:
```suggestion
    "uri_format": "https://data.4dnucleome.org/biosources/$1",
````

seems like the issue was just a missing comma

File: src/bioregistry/data/bioregistry.json [hunk 2]
```diff
@@ -96,7 +96,11 @@
...
+    "wikidata": {
+      "database": "Q110613621"
+
```
cthoyt, line 102:
```suggestion
```

cthoyt:
The issue was that the JSON was missing a comma, just be careful next time! I also suggest reading the feedback given by CI - it's there to automatically give feedback to you. you can also run `pip install tox` then `tox` to run the tests in a 100% reproducible way

PR #331, opened 2022-03-05
Improve attribution for external resources

### LINKED ISSUE #329: FAIRsharing Attribution
Greetings from FAIRsharing! 

I recently had a look at https://bioregistry.io/ now that our new site and API have been released, and I noticed that you have been successful at using our new API, that's great! Please let us know if you have any questions.

We have a couple of comments in regards to your site and how FAIRsharing is represented.

- Attribution - in documentation. Thank you very much for listing FAIRsharing in a few places on your website, and clearly stating within the JSON downloads those parts of your metadata come from us (e.g. https://bioregistry.io/api/registry/abs?format=json). To fully attribute us in compliance with our CC BY-SA 4.0 licence, we also request that you mention FAIRsharing as a source of your metadata within your documentation, for example in a "data sources" section (together with the other sources you pull information from) on your summary page (https://bioregistry.io/summary) and in the acknowledgements (https://bioregistry.io/acknowledgements). Full details on attributing and licencing are here: https://fairsharing.org/licence.
- Attribution - in metadata sourced from FAIRsharing. We request that any data taken from FAIRsharing, either by the API, scraping the site or other means, is attributed and cited appropriately. We request that a link and the FAIRsharing logo is presented next to the data taken from FAIRsharing. Whenever the description (or other metadata) comes from FAIRsharing, a link/logo to us should be present next to the relevant metadata.
- Metadata registry. You state that Bioregistry is "A collection of metadata about registries and mappings between the resources they describe...So far, the Bioregistry is the only metaregistry.". For your information, according to this definition, FAIRsharing is most definitely a metadata registry, as are a number of other resources such as the OpenAIRE knowledge graph.

Thank you for using FAIRsharing, it's great to see how the community makes use of our registry.

Best wishes,
Allyson

cthoyt:
Closes #329, this PR adds better attributions for external resources

- [x] Add `logo_url` to metaresource Pydantic model
- [x] Add logo to metaresource page, resource mapping
- [x] Add listing of external resources to acknowledgements section
- [x] Add specific FAIRsharing note to external curation page

Point by point discussion from #329:

> Attribution - in documentation. Thank you very much for listing FAIRsharing in a few places on your website, and clearly stating within the JSON downloads those parts of your metadata come from us (e.g. https://bioregistry.io/api/registry/abs?format=json). To fully attribute us in compliance with our CC BY-SA 4.0 licence, we also request that you mention FAIRsharing as a source of your metadata within your documentation, for example in a "data sources" section (together with the other sources you pull information from) on your summary page (https://bioregistry.io/summary) and in the acknowledgements (https://bioregistry.io/acknowledgements). Full details on attributing and licencing are here: https://fairsharing.org/licence.

- While the metaregistry listing itself (https://bioregistry.io/metaregistry/) reflects which external registries are ingested into the Bioregistry, we added a header on the acknowledgements page (https://bioregistry.io/acknowledgements) that lists all resources, their logos (if available), their licensing information (if available), and a link to the code that's used to ingest the external data.
- Currently the summary page is about the high-level problems the Bioregistry solves and not the external resources, so I don't think it's appropriate to include a link to any specific external resources here (besides the mention in the external coverage assessment). I added a sentence linking to the metaregistry listing there as a reminder that there's more information available elsewhere on the site.

> Attribution - in metadata sourced from FAIRsharing. We request that any data taken from FAIRsharing, either by the API, scraping the site or other means, is attributed and cited appropriately. We request that a link and the FAIRsharing logo is presented next to the data taken from FAIRsharing. Whenever the description (or other metadata) comes from FAIRsharing, a link/logo to us should be present next to the relevant metadata.

The only information from FAIRsharing that's displayed on the Bioregistry website are the FAIRsharing identifiers that have been mapped to other resources, so I don't think that a discussion about licensing is applicable. Otherwise, all content displayed on the Bioregistry site is either original or comes from other resources (specifically to avoid any licensing issues).

However, we liked the idea of adding images for each external registry and started to do so in this PR (see screenshot below of the mappings section of https://bioregistry.io/registry/chebi)

Screen Shot 2022-03-05 at 16 38 01

The information from FAIRsharing that's used for semi-automated curation of mappings at https://github.com/biopragmatics/bioregistry/tree/main/src/bioregistry/data/external/fairsharing now has an accompanying README that restates the FAIRsharing license.

> Metadata registry. You state that Bioregistry is "A collection of metadata about registries and mappings between the resources they describe...So far, the Bioregistry is the only metaregistry.". For your information, according to this definition, FAIRsharing is most definitely a metadata registry, as are a number of other resources such as the OpenAIRE knowledge graph.

There seems to be a bit of a misunderstanding about what the Bioregistry is doing and how it's naming its different faculties. Indeed there are many different kinds of registries and repositories of metadata, but we're only focusing on registries of prefixes for ontologies, terminologies, controlled vocabularies, and other semantic spaces in the biomedical domain. There's a difference between the term _metaregistry_ that we introduce here and metadata registry - the job of the metaregistry is to keep mappings between different registries of prefixes. For example, it has explicit curation of the fact that Identifiers.org uses `taxonomy` as the prefix for the NCBI Taxonomy and that OBO Foundry uses `NCBITaxon` as the prefix for the NCBI Taxonomy. Having this kind of information is crucial for mapping between different datasets created by people working in different contexts (e.g., SBML modelers often use Identifiers.org prefixes while ontologists often use OBO Foundry prefixes). Additionally, this functionality needs to be generalization and not just applicable to a small number of specific external registries.

This is very different from a general registry that contains metadata. I don't think it's the case that FAIRsharing explicitly curates these kinds of mappings in a generalizable way. Last time I checked, I noticed FAIRsharing does generate cross-links to BioPortal and the OBO Foundry for some terminologies, but it seemed to be rule-based (by string formatting) and made several mappings to resources that didn't exist. I'm curious to know if that's no longer the case with the new website, but either way I think we're talking about different things.


allysonlister:
Thank you for the updates to Bioregistry, and for the very clear explanations where I had misunderstood Bioregistry's purpose. Everything looks great, thank you.

To address your final point, I now understand what you mean by "metaregistry". You are correct in your understanding of how Bioregistry is unique, and how its purpose differs from a resource like FAIRsharing. We do store cross-references, but don't store how the prefixes etc are different across resources. We do try to keep on top of broken links (but some additional ones have crept in as we focused on the release of the new site over the last few months that we are actively working on), so if you see a systemic or a particular issue, please feel free to let us know, either at contact@fairsharing.org or at https://github.com/FAIRsharing/fairsharing.github.io/issues

Thanks again, and best wishes!

PR #335, opened 2022-03-09
Update GitHub pattern

cthoyt:
This PR changes the github pattern such that it can accept both users/organizations and also repositories. Suggested by @egonw


egonw:
Yes, I think this will do fine.

egonw:
cc @mkutmon

PR #343, opened 2022-03-11
Add Wikidata database link for `abcd` and `abm`

CondeArcana333:
add wikidata database to abcd and abm

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -194,11 +194,7 @@
     "mappings": {
       "cellosaurus": "ABM"
     },
-    "name": "Applied Biological Materials cell line products",
-    "synonyms": [
-      "abm"
-    ],
-    "uri_format": "https://www.abmgood.com/Cell-Biology.html$1",
+   
     "wikidata": {
       "database": "Q111191868"
     }
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -177,7 +177,10 @@
       "prefix": "ABCD",
       "uri_format": "https://web.expasy.org/cgi-bin/abcd/search_abcd.pl?input=$1"
     },
-    "uri_format": "https://web.expasy.org/abcd/ABCD_$1"
+    "uri_format": "https://web.expasy.org/abcd/ABCD_$1",
+    "wikidata": {
+      "database": "Q110998933"
+    }
   },
   "abm": {
     "cellosaurus": {
@@ -190,6 +193,14 @@
     "example": "T0599",
     "mappings": {
       "cellosaurus": "ABM"
+    },
+    "name": "Applied Biological Materials cell line products",
+    "synonyms": [
+      "abm"
+    ],
+    "uri_format": "https://www.abmgood.com/Cell-Biology.html$1",
+    "wikidata": {
+      "database": "Q111191868"
     }
   },
   "abs": {


src/bioregistry/data/bioregistry.json
```diff
@@ -190,6 +193,14 @@
     "example": "T0599",
     "mappings": {
       "cellosaurus": "ABM"
+    },
+    "name": "Applied Biological Materials cell line products",
+    "synonyms": [
+      "abm"
+    ],
+    "uri_format": "https://www.abmgood.com/Cell-Biology.html$1",
```
cthoyt, line 201: This doesn't seem to work with an example, what's the motivation for adding this?

CondeArcana333: good afternoon, I only added it so that it was similar to the one I uploaded previously, in that they told me that it was good and I tried to make it as similar as possible, but if you prefer I will remove it.

cthoyt: The point of the URI format string is to make it possible to resolve a given local identifier - this is not the same as a homepage (perhaps there was some confusion here - https://cthoyt.com/2021/10/07/biopragmatics-glossary.html provides some more background). In a URI format string, the `$1` is a placeholder for the local unique identifier. You can see that there's already one of these available via Cellosaurus for this record.

src/bioregistry/data/bioregistry.json
```diff
@@ -190,6 +193,14 @@
     "example": "T0599",
     "mappings": {
       "cellosaurus": "ABM"
+    },
+    "name": "Applied Biological Materials cell line products",
+    "synonyms": [
+      "abm"
```
cthoyt, line 199: this synonym is redundant of the prefix, so it can be removed.

src/bioregistry/data/bioregistry.json
```diff
@@ -190,6 +193,14 @@
     "example": "T0599",
     "mappings": {
       "cellosaurus": "ABM"
+    },
+    "name": "Applied Biological Materials cell line products",
```
cthoyt, line 197: This seems to duplicate the name that's already available via cellosaurus, so it can be removed.

cthoyt:
Note that some of the issues I raised might be because it's not obvious that some entries draw from potentially several external registries. There are very extensive metadata quality checks so there's nothing missing that's required, however some fields are indeed optional that we could benefit from adding (such as URI format strings, in many cases). Please don't be discouraged by my comments - having a second eye checking metadata is always welcome! Could you also please share your ORCID identifier so I can show you how to add yourself as a contributor to the record?

CondeArcana333:
don't worry, now i delete the mentioned lines, i just leave the id from wikidata. And in one more moment I provide you with my ORCID

CondeArcana333:
I think I already added a commit correcting the previous one. Please let me know if anything else needs to be corrected.

cthoyt:
It looks like there's a spurious new line introduced in this PR. I wrote a command that helps keep this JSON file tidy - first you have to pip install the bioregistry in editable mode with `python -m pip install -e .` then you can run `bioregistry lint`

CondeArcana333:
I sent the orcid the @lmichan (layla michan) 0000-0002-5798-662X she is in charge of the laboratory, and in a moment I look for the line that causes the conflict to correct it.

lmichan:
Hola Charles, todo lo que estamos trabajando para Bioregistry lo hacemos entre cuatro personas, solamente ves a Adrían porque él está probando subir los registros en JSON a Bioregistry pero en realidad este es un trabajo en equipo, es importante mencionar que Victor, Adrían y  Minerva son voluntarios que estudian matemáticas y están aprendiendo en mi laboratorio aprendiendo sobre bases de datos biológicas. En realidad todos están haciendo un gran trabajo!

CondeArcana333:
@cthoyt Sorry, I was reviewing the code and I didn't find the problem, I don't know if you can help me correct it and I'll check what I do so I don't make that mistake again, or if it has already been corrected, or I could delete this pull request and do it from scratch. principle to correct the problem.

cthoyt:
Technically you can invite extra people as contributors to one of your forks (see [these instructions](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-user-account/managing-access-to-your-personal-repositories/inviting-collaborators-to-a-personal-repository)). Each entry in the Bioregistry also has a field called "contributors_extra" where you can list potentially several people. For example, https://bioregistry.io/registry/ncats.drug lists a few people who have edited this record. Also, to credit people, we need to have their ORCID identifiers (from https://orcid.org).

PR #352, opened 2022-03-18
Enable curation of prescriptive contexts

### LINKED ISSUE #324: Use case: maintaining a prescriptive context
I want to be sure while I am promoting bioregistry around that the following use case can be captured. 

We should be able to:

- [ ] define a _project specific context_, i.e. obo context
- [ ] associate some orcids with that context that declare us as the _maintainers_
- [ ] have some kind of _default order for which prefix sources_ (prefix commons, OBO etc) we want to prefer, and the rest being suggested as "defaults" by bioregistry
- [ ] _define the preferred prefixes for each URI expansion_ (example: we want to use `DOI:` for dois, not `doi:`)
- [ ] _define the preferred URI expansions for each prefix_ (example: we want to expand `DOI:` to "https://doi.org/", not any other URL)
- [ ] we need a mechanism to _fix preferred prefixes_ for each URI expansion (i.e. when we say: "preferred" somewhere and no matter what happens, we need to be able to rely on the fact that this preferred prefix cannot be changed without the explicit agreement of the maintainers documented above.
- [ ] we need a mechanism to _fix the preferred URI expansions_ for each prefix (i.e. we say: "preferred" somewhere and no matter what happens, we need to be able to rely on the fact that this preferred prefix cannot be changed without the explicit agreement of the maintainers documented above.

Can we confirm whether this extensive set of requirements can be met by bioregistry or not? Thanks :)

cc @dosumis @cmungall 

cf. https://obo-communitygroup.slack.com/archives/C01R2D66249/p1645718367637119

cthoyt:
Closes #324. This PR does the following:

- [x] Extends the existing notion of generating an OBO-specific context (exported as both JSON-LD and SHACL) to be configured from a JSON file
- [x] Adds PyDantic models and a JSON schema for this file
- [x] Adds unit tests for data integrity
- [x] Adds governance documentation (see text [here](https://github.com/biopragmatics/bioregistry/tree/contexts/exports/contexts#prescriptive-context-governance))
- [x] Adds web listing of contexts


src/bioregistry/data/contexts.json
```diff
@@ -0,0 +1,34 @@
+{
+  "obo": {
+    "custom_prefix_map": {
+      "DOI": "https://doi.org"
+    },
+    "description": "The OBO Context exists for the following reasons: ...",
```
cthoyt, line 6: @matentzn this needs to be updated before we merge this PR

exports/contexts/README.md
```diff
@@ -21,3 +21,92 @@ the `bioregistry.export.prefix_maps` module.
...
+  tracker, but only a pull request counts as a proposal.
+- New contexts must pass all CI tests before being considered.
+- The Bioregistry Review Team is responsible for reviewing all new context
+  requests and accepts at their discretion.
+- Each context must have a unique key. Similar rules to choosing a good prefix
+  in the Bioregistry apply.
+- Each context must list its maintainers along with their current GitHub, ORCID,
+  and optional email.
+- Each context must include a name and description
+- Contexts should ideally be significantly different from the Bioregistry
```
matentzn, line 52: This does not need to be here I think, but doesn't hurt. I think there is a point to be made especially for contexts that are almost the same as bioregistry as this encourages gravitation to using the defaults.

src/bioregistry/data/contexts.json
matentzn: 1. Reconciling "database cross-references" used by OBO ontologies and centralising their validation.
2. Standardising and validating SSSOM mappings provided in the context of OBO ontologies.
3. Providing an incentive for the documentation of non-standard prefixes for OBO ontologies.

matentzn (commented):
This is amazing. Excellent governance processes! We need these for like everything we do in any repo.

exports/contexts/README.md
cthoyt: Alright good point, if we don't discourage people from doing questionable things, then we can catch the ones who are doing it and try and make this case to them on an _ad-hoc_ basis

src/bioregistry/data/contexts.json
cthoyt: Added in ff02f14

exports/contexts/README.md
cthoyt: Removed in ff02f14

PR #379, opened 2022-05-17
Add prefix parking guidelines

### LINKED ISSUE #365: Policies on prefix parking
Let's start with a bioregisrty definition of **substantiation**: this is when it's provable that a semantic space exists by one or more of the following:

1. There's a public place where you can get the entire list of terms. Ontologies (e.g., GO) and database (e.g., HGNC) usually make this pretty straightforwards
2. There's a working, public URI format string that either lets you get HTML, JSON, RDF, or some other kind of information artifact for a given local unique identifier. For example, OMIM.PS doesn't have a way to get a full list of terms but if you have a given identifier you can use it with the URI format string to see

A slightly less strict criteria for "substantiation" could be that references to a semantic space can be found in public external resources or ontologies. This happens quite a lot and many entries in the Bioregistry have been added based on this (e.g., OpenCyc references pop up quite frequently but this resource hasn't existed for more than a decade).

This can help us define **parking** which is when a prefix is registered for a semantic space that does not yet exist or is currently in development (which by definition, means it isn't yet substantiated). This issue is the starting point for some policies regarding prefix parking, which was in part motivated by the discussion on #359. Here are some thoughts on policies for how to consider prefix parking:

1. While it's not typically the job of the Bioregistry Review Team to cast judgement on the motivation for why a resource exists or whether it should exist at all, to protect the integrity and quality of the Bioregistry, in the case when a group wants to park a prefix, that group must seek out and seriously consider suggestions from the Bioregistry Review Team - our business is prefixes, CURIEs, URIs, etc., so we know what we're talking about more than most submitters. New prefix requests from groups unwilling to seriously consider suggestions will be dismissed and asked to re-submit when their prefixes are substantiated.
2. Parked prefixes that are not substantiated in a certain period of time will be subject to removal (e.g., in case someone else wants to use that prefix - the fact that it has not been substantiated, again, by definition, means this will have zero impact)
3. It's not the job of the Bioregistry to support parking prefixes for semantic spaces that will not be public or that won't be used in other public resources, these can be rejected.

cthoyt:
Closes #365

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -142,6 +142,66 @@ individual prefix requests, which each get their own discussion thread, pull
 request, and CI/CD runs. Submitters of bulk prefix requests that contain many
 issues may be asked to re-submit as individual prefix requests.
 
+#### Prefix Parking
+
+A prefix and its corresponding semantic space are **substantiated** when it's
+provable that a semantic space exists by one or more of the following:
+
+1. There's a public place where you can get the entire list of terms.
+   Ontologies (e.g., [Gene Ontology](https://bioregistry.io/go)) and databases
+   (e.g., [HGNC](https://bioregistry.io/hgnc)) usually make this pretty
+   straightforwards by offering download links for the ontology or full
+   database.
+2. There's a working, public URI format string that either lets you get HTML,
+   JSON, RDF, or some other kind of information artifact for a given local
+   unique identifier. For example, [OMIM.PS](https://bioregistry.io/omim.ps)
+   doesn't have a way to get a full list of terms but if you have a given local
+   unique identifier, you can use it with the URI format string to retrieve some
+   information about the entity corresponding to that LUI.
+
+A potential less strict third criteria for substantiation could be when
+references to entities in a semantic space (i.e., in the form of CURIEs) can be
+found in public resources or ontologies that are external to the resource in
+which the prefix/semantic space are defined. This is more common for historical
+prefixes (e.g., OpenCyc references appear quite frequently, but this resource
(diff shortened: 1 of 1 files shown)


docs/CONTRIBUTING.md
```diff
@@ -142,6 +142,69 @@ individual prefix requests, which each get their own discussion thread, pull
...
+https://github.com/biopragmatics/bioregistry/issues/359.
+
+1. While it's not typically under the purview of the Bioregistry Review Team to
+   judge the utility of a prefix nor comment on its corresponding design
+   decisions (e.g., choice of local unique identifier scheme, regular expression
+   pattern, URI format string), submitters seeking to park a prefix must both
+   actively publicly seek out and seriously consider suggestions and advice from
+   the Bioregistry Review Team with regards to these matters (e.g., in the issue
+   corresponding to a new prefix request). Submissions unable/unwilling to
+   follow these guidelines will be dismissed and asked to re-submit after their
```
sierra-moxon, line 186: @cthoyt - would a caveat here be that the bioregistry review team would be willing to consider "justifications" provided by the submitter that might convince them to bend the best practices?  I know this sounds like it makes the good practices invalid, but we're in the domain of exceptions and we do want people to use bioregistry with preference, but they might not always be able to maneuver into the guidelines in the identifiers publication, etc. I guess what I am asking is for some wiggle room here (provided there is a "good reason" for that wiggle room).

cthoyt: I will soften this language - and leave it up to the Bioregistry review team on a case by case basis.

```suggestion
   follow these guidelines may be dismissed and asked to re-submit after their
```

PR #391, opened 2022-05-24
Adding automated release

Adafede:
@cthoyt 

I guess this should do the trick. 

Btw, I was not able to test it on my fork because of the local changes...

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,14 @@
+name: Release
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: marvinpinto/action-automatic-releases@v1.2.1
+        with:
+          repo_token: "${{ secrets.GITHUB_TOKEN }}"
+          automatic_release_tag: "latest"
+          prerelease: false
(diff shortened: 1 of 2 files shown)


cthoyt:
The inaugural run of the action: https://github.com/biopragmatics/bioregistry/actions/runs/2378040486

cthoyt:
Seems like everything worked alright except it named it "latest" instead of mirroring the tag name as the release title

.github/workflows/release.yml
```diff
@@ -0,0 +1,14 @@
...
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: marvinpinto/action-automatic-releases@v1.2.1
+        with:
+          repo_token: "${{ secrets.GITHUB_TOKEN }}"
+          automatic_release_tag: "latest"
```
cthoyt, line 13: can we change this bit to use the same name as the tag?

Adafede:
Oh crap!

We should just remove  https://github.com/biopragmatics/bioregistry/blob/7d6f389d59a90ba4fa65a7a6870ffbda43b338a2/.github/workflows/release.yml#L13

competely and should be OK

.github/workflows/release.yml
Adafede: Sorry as the PR was closed made a new one with the requested change

PR #412, opened 2022-06-06
Update release.yml

Adafede:
Adding up to 4 digits to the pattern for automatic release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@ name: Release
 on:
   push:
     tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+'
+      - 'v[0-9]{1-4}+.[0-9]{1-4}+.[0-9]{1-4}+'
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -2,7 +2,7 @@ name: Release
 on:
   push:
     tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+'
+      - 'v[0-9]{1-4}+.[0-9]{1-4}+.[0-9]{1-4}+'
```
cthoyt, line 5: I'm a bit confused if how this interacts with the plus, and why not just use `\d` instead of `[0-9]`?

Adafede:
Not sure it is more readable but you're right! Updated

.github/workflows/release.yml
```diff
@@ -2,7 +2,7 @@ name: Release
 on:
   push:
     tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+'
+      - 'v\d+\.\d+\.\d+'
```
cthoyt, line 5: I was reading https://stackoverflow.com/questions/61891328/trigger-github-action-only-on-new-tags and wonder if there's any reason to specify this pattern at all, and not just use `'*'`?

PR #432, opened 2022-06-22
Add prefix: phosphosite.sitegroup

cthoyt:
As a follow-up to [this slack discussion](https://syspharm.slack.com/archives/D01QC9NDUCS/p1655916778549459), this PR adds a prefix for PhosphoSite Site Groups. Note, I couldn't figure out what `protOrg=1156` made different from other integers, so this is included in the URI format string.


bgyori (commented):
Unfortunately this URL scheme doesn't work in general. For example, https://www.phosphosite.org/siteGroupAction.action?id=964500638&protOrg=3564273&showAllSites=true works but https://www.phosphosite.org/siteGroupAction.action?id=964500638&protOrg=1156&showAllSites=true doesn't. Perhaps we could keep the record but without a resolver/url.

cthoyt:
Hmm that is indeed upsetting. Luckily records automatically link back to github discussions so if we want to look into this more we won't have to remember where the discussion is :)

PR #443, opened 2022-07-08
Update bioregistry.json

megbalk:
Using the noaa prefix as an example, I want to see if I did this correctly.

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -68320,6 +68320,13 @@
       "name": "Zebrafish Phenotype Ontology",
       "prefix": "ZP"
     },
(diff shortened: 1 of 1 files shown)


cthoyt:
There are a couple issues here - you actually need to add a top-level entry that has "name", "homepage", "description" in it. Maybe a simpler way is to use the form at https://github.com/biopragmatics/bioregistry/issues/new?assignees=biopragmatics%2Fbioregistry-reviewers&labels=New%2CPrefix&template=new-prefix.yml&title=Add+prefix+%5BX%5D

cthoyt:
Closing in favor of #462, which was auto-generated based on https://github.com/biopragmatics/bioregistry/issues/456

PR #457, opened 2022-07-13
Implement getting pattern with namespace embedded

bgyori:
This PR adds a method for the `Resource` class called `get_pattern_re_with_banana` which returns a version of the identifier pattern that matches the embedded namespace (aka banana) in the identifier.

Some example usages:
```python
In [1]: import bioregistry
In [2]: ch = bioregistry.get_resource('chebi')

In [3]: ch.get_pattern_re_with_banana()
Out[3]: re.compile(r'^(CHEBI:)\d+$', re.UNICODE)

In [4]: ch.get_pattern_re_with_banana().match('1234')

In [5]: ch.get_pattern_re_with_banana().match('CHEBI:1234')
Out[5]: <re.Match object; span=(0, 10), match='CHEBI:1234'>

In [6]: ch.get_pattern_re_with_banana(strict=False).match('1234')
Out[6]: <re.Match object; span=(0, 4), match='1234'>

In [7]: ch.get_pattern_re_with_banana(strict=False).match('CHEBI:1234')
Out[7]: <re.Match object; span=(0, 10), match='CHEBI:1234'>
```

### DIFF
--- a/src/bioregistry/schema/struct.py
+++ b/src/bioregistry/schema/struct.py
@@ -32,9 +32,9 @@ from bioregistry.schema.utils import EMAIL_RE
 from bioregistry.utils import removeprefix, removesuffix
 
 try:
-    from typing import Literal, Match  # type:ignore
+    from typing import Literal, Pattern  # type:ignore
 except ImportError:
-    from typing_extensions import Literal, Match  # type:ignore
+    from typing_extensions import Literal, Pattern  # type:ignore
 
 __all__ = [
     "Attributable",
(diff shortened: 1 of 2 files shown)


cthoyt:
@bgyori do you think it would be useful to assign a capture group to the actual local unique identifier?

src/bioregistry/schema/struct.py
```diff
@@ -669,6 +669,78 @@ def get_pattern_re(self):
...
+        if not banana:
+            return pattern
+
+        banana_peel = self.get_banana_peel()
+        prepattern = f"{banana}{banana_peel}"
+        if not strict:
+            prepattern = f"({prepattern})?"
+        return "^" + prepattern + pattern.lstrip("^")
+
+    def get_pattern_re_with_banana(self, strict: bool = True):
```
bgyori, line 709: I'm curious, is it problematic to add the `re.Pattern` annotation for the return value?

cthoyt: It’s changed so much since py37 that idk a way to do it without writing a compatibility layer

PR #484, opened 2022-07-29
Added flag `with_preferred_prefix` to `curie_from_iri`

### LINKED ISSUE #482: Is there a reason why `curie_from_iri` returns a normalized prefix rather than the preferred prefix?
When I run the following:

```python
from bioregistry import curie_from_iri

curie_from_iri("http://purl.obolibrary.org/obo/GO_9999999")
```
gives me `go:9999999` but I want `GO:9999999` which would be a CURIE with a 'preferred prefix'. As of now I have massaged the data to get it as follows:

```python
from bioregistry import parse_iri, get_preferred_prefix, curie_to_str

prefix, identifier = parse_iri("http://purl.obolibrary.org/obo/GO_9999999")
preferred_prefix = get_preferred_prefix(prefix)
curie = curie_to_str(preferred_prefix, identifier)
```

Questions:

1.  Is there a reason why the preferred prefix isn't the default result of `parse_iri`
2. Is there a more elegant way of doing this?
3. If I provide a prefix map, it still gives me the normalized prefix. Although I'm not sure how that would help either. Example below:
```python
from bioregistry import curie_from_iri

curie_from_iri(
    "http://purl.obolibrary.org/obo/GO_9999999", 
    prefix_map={"GO": "http://purl.obolibrary.org/obo/GO_"},
)
```

cc: @matentzn @sierra-moxon (fyi)

hrshdhgd:
This fixes #482 

Added a flag `with_preferred_prefix` to `curie_from_iri` which defaults to `False`.

What does it do?
If set to `True`, the function `curie_from_iri` returns a CURIE with the preferred prefix (if one exists). By default, it returns the normalized prefix.

### DIFF
--- a/src/bioregistry/parse_iri.py
+++ b/src/bioregistry/parse_iri.py
@@ -5,7 +5,7 @@
 from functools import lru_cache
 from typing import List, Mapping, Optional, Tuple, Union
 
-from .resolve import parse_curie
+from .resolve import get_preferred_prefix, parse_curie
 from .resource_manager import prepare_prefix_list
 from .uri_format import get_prefix_map
 from .utils import curie_to_str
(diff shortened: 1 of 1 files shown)


cthoyt:
@hrshdhgd thanks for looking into this. See also #483 where I did something similar and made several other unifications

Issue #491, opened 2022-08-03
Perfecting the project-specific prefix maps


matentzn:
I am looking at the latest update of the OBO context and have a few questions.

## Replaced OBO PURLs when source ontologies are obsolete

Is this a general design decision? Not to include OBO prefixes of deprecated sources at all? Or if they conflict with something else?
 
```
"ADW": "http://purl.obolibrary.org/obo/ADW_",
"ADW": "https://animaldiversity.org/accounts/",
"IP": "http://purl.obolibrary.org/obo/IPR_",
"IP": "https://www.ebi.ac.uk/interpro/entry/",
```

## Adding some weird sounding prefixes.. Why?

```
"Geomames": "https://www.geonames.org/", (typo?)
"Geonamaes": "https://www.geonames.org/", (typo?)
"IC9CM": "http://icd9cm.chrisendres.com/index.php?action=search&srchtext=", (missing D?)
"NCBI Taxonomy": "http://purl.obolibrary.org/obo/NCBITaxon_", (prefixes with spaces?)
```

## Shouldnt this context be a source-unique bi-map?

There are many alternative ICD10 (and SNOMEDCT) prefixes in our obo context, lower case, upper case, variations. I would only want 1! For URI to CURIE conversions, how will my tools understand which prefix to pick (based on the context alone)? This is what we have:

```
"IC9CM": "http://icd9cm.chrisendres.com/index.php?action=search&srchtext=",
"ICD": "http://apps.who.int/classifications/icd10/browse/2010/en#/",
"ICD-10": "http://apps.who.int/classifications/icd10/browse/2010/en#/",
"ICD-11": "https://icd.who.int/browse11/l-m/en#/http://id.who.int/icd/entity/",
"ICD10": "http://apps.who.int/classifications/icd10/browse/2010/en#/",
"ICD10CM": "https://icd.codes/icd10cm/",
"ICD9": "http://www.icd9data.com/getICD9Code.ashx?icd9=",
"ICD9CM": "http://icd9cm.chrisendres.com/index.php?action=search&srchtext=",
"ICD9CM_2005": "http://icd9cm.chrisendres.com/index.php?action=search&srchtext=",
"ICD9CM_2006": "http://icd9cm.chrisendres.com/index.php?action=search&srchtext=",
"ICD9_2005": "http://www.icd9data.com/getICD9Code.ashx?icd9=",
[...]

PR #494, opened 2022-08-04
Add blacklist to contexts

cthoyt:
Related to #491, needs review from @matentzn


matentzn (approved):
I didn't review the python code, just the behaviour. I think this is great.

cthoyt:
@matentzn can you suggest any other blacklist prefixes for the OBO context?

matentzn:
icd10, icd9, are the only two I can see at a quick glance I don't want people to use.. (they should us icd10cm, icd9cm)

PR #509, opened 2022-08-18
Add curations based on KGOBO dump

cthoyt:
This PR adds various curations based on the KGOBO dump - it should handle a large number of the missing nodes

CC @LucaCappelletti94

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -27177,6 +27177,13 @@
         "homepage": "https://gnome.glyomics.org/StructureBrowser.html",
         "name": "GNOme Structure Viewer",
         "uri_format": "https://gnome.glyomics.org/StructureBrowser.html?focus=$1"
+      },
+      {
+        "code": "gno.composition",
+        "description": "GNOme composition browser",
+        "homepage": "https://gnome.glyomics.org/CompositionBrowser.html",
+        "name": "GNOme Composition Browser",
+        "uri_format": "https://gnome.glyomics.org/CompositionBrowser.html?focus=$1"
       }
     ],
     "synonyms": [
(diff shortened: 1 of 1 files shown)


LucaCappelletti94:
The efforts made by [Harry Caufield](https://github.com/caufieldjh) are symmetrical to these, so likely some coordination could help achieve complete standardization faster: https://github.com/Knowledge-Graph-Hub/kg-obo/issues/171#issuecomment-1218234564

If it could be of help, I can rerun the same script from yesterday using the bioregistry from this pull request.

cthoyt:
Thanks for linking to that! @caufieldjh you're always welcome to make issues to improve bioregistry as well :)

After this PR is done I will make a new release and Luca can re-run his analysis, potentially adding a few extra rules to throw away certain content that won't be helpful (maybe also informing improvement to generation of KGOBO, like throwing away any content from Fraunhofer ;))

LucaCappelletti94:
One more thing: I can also run bioregistry on Monarch and KGHUB, which I have not done so far. Should I wait for the new release?

LucaCappelletti94:
Oh, and @callahantiff PheKnowLator may be likely parsable as IRIs, though the are presented as `<IRI>` if I am not mistaken.

cthoyt:
There are also quite a few new prefix requests I made (https://github.com/biopragmatics/bioregistry/issues?q=is%3Aissue+is%3Aopen+label%3ANew) while working through this, I asked various @biopragmatics/bioregistry-reviewers to take a look at these so I don't want to rush them. Anyway, it would be good to generalize this script so we can re-use it and report which version of bioregistry was used for each so we can see improvements over time

cthoyt:
@LucaCappelletti94 maybe if you add some additional tooling for grouping URIs with similar beginnings it would be helpful for review

LucaCappelletti94:
Could you further detail what that may look like? Like, a very short example?

caufieldjh:
> Thanks for linking to that! @caufieldjh you're always welcome to make issues to improve bioregistry as well :)

After this PR is done I will make a new release and Luca can re-run his analysis, potentially adding a few extra rules to throw away certain content that won't be helpful (maybe also informing improvement to generation of KGOBO, like throwing away any content from Fraunhofer ;))

Sure - I'm planning to add normalization to Bioregistry prefixes imminently.

callahantiff:
> Oh, and @callahantiff PheKnowLator may be likely parsable as IRIs, though the are presented as `<IRI>` if I am not mistaken.

Thanks @LucaCappelletti94 -- the next release of PheKnowLator will be normalized to Bioregistry too. 🎉
I am also add some additional output file types, the type you mentioned is serialized NT, which is why there are brackets.

Issue #485, opened 2022-07-29
Add prefix ICD10WHO


joeflack4:
### Prefix

ICD10WHO

### Name

World Health Organization

### Homepage

https://icd.who.int/browse10/2019/en#/

### Description

ICD10 is well known. There exists a WHO version, which is essentially a base version from which other versions are derived. However, ICD10 is often mistaken to mean ICD10CM. Thus, to avoid ambiguity, we propose ICD10WHO as a community driven prefix.

### Example Local Unique Identifier

C34

### URI Format String

https://icd.who.int/browse10/2019/en#/$1

### Contributor Name

Joe Flack

### Contributor GitHub

joeflack4

### Contributor ORCiD

0000-0002-2906-7319

### Additional Comments

See discussion: https://github.com/biopragmatics/bioregistry/issues/251

cthoyt:
@joeflack4 we'll need a few things to move forwards:

1. Additional example local unique identifies to get a better feel of what's in this semantic space
2. Regular expression for *all* terms in this semantic space
3. An updated description - this should be for people who don't know what ICD10WHO is. This should _not_ be a justification of project-specific usage, but rather text to let people who aren't familiar with this resource to understand what it is. Ideally, this also includes a description of how it's different from ICD10 and ICD10CM, and when it's appropriate to use it instead

cthoyt:
@matentzn so the other option was to rename the existing `icd10` to `icd10who` -> is that an alternative?

matentzn:
I think `icd10` is like `dc` - you cant tell every what it refers to. Is it possible to add `icd10`, and `ICD10`, as deprecated prefixes to both icd10cm and icd10who? Since it can be both, we should discourage using it altogether.

matentzn:
> 1. Additional example local unique identifies to get a better feel of what's in this semantic space

So many different things..:

Examples of grouping classes in ICD10WHO:
- [V20-V29](https://icd.who.int/browse10/2019/en#/V20-V29) Motorcycle rider injured in transport accident
- [G00-G09](https://icd.who.int/browse10/2019/en#/G00-G09) Inflammatory diseases of the central nervous system

Examples of leaf codes:

- V22: Motorcycle rider injured in collision with two- or three-wheeled motor vehicle
- D01: Carcinoma in situ of other and unspecified digestive organs

> 2. Regular expression for all terms in this semantic space

[...]

Issue #251, opened 2021-11-17
ICD9/ICD10: can we determine a good canonical URI prefix?


matentzn:
We were wondering what would be a suitable canonical URI prefix for ICD-9 codes. Does anyone have any suggestions/opinions on that? @dhimmel @cthoyt 

I guess the main tradeoff to consider here are:

- degree of officiality
- resolvability, ie.g. https://icd.codes/icd10cm/H108, https://icd.codes/icd9cm/3007

is https://icd.codes/... suitable?

sierra-moxon:
I took an ICD9 dbxref from mondo: ICD9:784.7 and tried to resolve it using the links above, and I can't.  I can however find a page that redirects ICD9:784 to the ICD10 equivalent.  https://icd.codes/icd9cm/784 -- and then in the synonyms list for ICD9:784 I see 784.7 which is linked like this: https://icd.codes/icd9cm/7847

sierra-moxon:
I also have some references to ICD0 (but can't find a resolvable identifier for that).

dhimmel:
For the record, there currently isn't any revolvers for ICD9 in the Bioregistry as per https://bioregistry.io/reference/icd9:277.9:

> ![image](https://user-images.githubusercontent.com/1117703/142640345-fba6997c-59d3-4460-ba1a-fc134b3a0e43.png)

[For ICD10](https://bioregistry.io/reference/icd10:C34), the Bioregistry uses resolution URLs like http://apps.who.int/classifications/icd10/browse/2010/en#/C34. This URL is by the World Health Organization, which creates ICD codes. But the WHO doesn't provide a browser for ICD9 codes, right? And that is why we need to pick a secondary provider?

https://icd.codes looks good. Would prefer if its source code was open. Probably better to put something to resolve ICD9 codes than nothing.

joeflack4:
I can't remember if it was ICD10 or ICD11, but Chris Chute was explaining to me that there is a base "ICD", which I think might be a polyhierarchy, and then they go through "linearization" to make mono-hierarchies ICDxxCM and ICDxxPS. If I find the GitHub thread I'll link it. I also don't know if this applies to just ICD10, just ICD11, or both ICD10 and ICD11. And I don't know if this also applies to pre-ICD10.

matentzn:
[...]

PR #523, opened 2022-08-22
Add ICD10WHO as synonym for ICD10

cthoyt:
Based on discussion with Chris Chute (https://github.com/biopragmatics/bioregistry/issues/251#issuecomment-1097994811), these are the same. Alternative to #485/#486.

@matentzn @joeflack4 thoughts?

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -31782,7 +31782,8 @@
     "synonyms": [
       "ICD",
       "ICD-10",
-      "ICD10"
+      "ICD10",
+      "ICD10WHO"
     ],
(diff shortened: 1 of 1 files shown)


matentzn:
I am ok, I really would prefer to discourage the use of ICD10 altogether, in favour of the specific alternatives. In the end, we will have to make a decision that contradicts some of the current practice, but that is ok.

cthoyt:
@matentzn I am all about making practical decisions that upend current practice! But this ICD thing has been such a pain to figure out since all of the answers to my questions have been very fluffy and not super actionable. I will now close all of these discussions and make a new unifying thread to move forwards with concrete points :) thank you and @joeflack4 for all of the help so far, and I'm counting on you going forwards!

PR #524, opened 2022-08-22
Use ICD10WHO in OBO context

cthoyt:
Related to #491

@matentzn this updates the OBO context and needs your approval - it might actually finish the issue #491 completely. Please let me know


matentzn (approved):
Great, thank you!

cthoyt:
@matentzn btw should we also map the `geogeo` prefix to `GEO` and `geo` to something else?

PR #540, opened 2022-08-31
Added another 4 WikiPathways articles

egonw:


src/bioregistry/data/bioregistry.json
```diff
@@ -72592,6 +72592,30 @@
         "pmc": "PMC2475545",
         "pubmed": "18651794",
         "title": "WikiPathways: pathway editing for the people."
+      },
+      {
+        "doi": "10.1093/NAR/GKAA1024",
```
cthoyt, line 72597: seems like these are case insensitive to the DOI resolver but the article page shows it in lowercase so we should probably stick to that

egonw: I learned from the Wikidata project that DOIs are by design case insensitive

cthoyt: Interesting, so if it's a free-for-all, then we'll require they're standardized to lowercase in the Bioregistry.

egonw: I can live with that. I'll make an update commit.

src/bioregistry/data/bioregistry.json
```diff
@@ -74466,4 +74490,4 @@
     },
     "pattern": "^\\d+$"
   }
-}
\ No newline at end of file
+}
```
cthoyt, line 74493: sorry to say but you have to remove the trailing newline otherwise the lint check won't pass. Besides that, this is ready to merge

cthoyt (approved):

src/bioregistry/data/bioregistry.json
egonw: wow... I thought we had left windows/linux/mac newline issues years behind us.

cthoyt:
Thanks @egonw! Glad to have more high-quality metadata in the Bioregistry 🚀

PR #542, opened 2022-09-01
Update Wikipathways description and license

egonw:
Per suggestion here: https://github.com/identifiers-org/identifiers-org.github.io/issues/205

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -72563,6 +72563,7 @@
         "orcid": "0000-0001-7542-0286"
       }
     ],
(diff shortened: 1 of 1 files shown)


egonw:
(sorry, stupid Kate...)

cthoyt:
However, before merging, I'd suggest improving the descriptive text here further.

egonw:
> note that we store the MIRIAM data but do not update it ourselves. 

Yes, that's what I meant when I said would it not be overwritten :( and why I had not created that pr until you asked :(

cthoyt:
No problem though, check the diff for the solution!

PR #553, opened 2022-09-03
Improve URI parsing with tries

### LINKED ISSUE #437: Improve URI parser speed with `pytrie`
Right now, the URI parser is pretty slow since it's linear with the number of URI format strings. This was improved in #481, but perhaps it could be improved even further using a more appropriate data structure and algorithm (memory trade-off isn't a big issue here since the Bioregistry is only a few thousand data)

See https://github.com/gsakkis/pytrie for a data structure that can take this sub-linar

cthoyt:
Closes #437

1. Use `curies` for maximum speed (4.7K/s to 124.7K/s)
2. Generate a more thorough reverse prefix map
   - Required _tons_ of curation of duplicates which I wasn't aware of
   - Motivates even further data model changes since getting externally mapped data is a pain

## Depends on

- [x] #552 
- [x] #551


cthoyt:
@lucaCappelletti94 you'll be happy to know this improves parse speeds 25-30x.

LucaCappelletti94:
I'm a bit under the weather, so it's particularly nice to hear news such as these!

cthoyt:
same here, been sick since monday :/

PR #577, opened 2022-09-19
Update scop

cthoyt:
When I was going to do some curation of EDAM terms, I noticed that the SCOP entry actually was pointing to two different websites (albeit with similar names) that have completely different identifier schemata and data. This PR splits them up.

I made a screencast of when I was looking into this PR and put it on YouTube here https://www.youtube.com/watch?v=zUKUSm0C3Z0


cmungall:
@jmchandonia thoughts?

megbalk:
@cthoyt I'm away this week and won't be able to review until next week

cthoyt:
FYI I made a video about the whole though process behind this and put it on youtube https://www.youtube.com/watch?v=zUKUSm0C3Z0

jmchandonia:
Thanks @cthoyt - your video was very helpful in helping to figure out your thought process.

The complication is that there are actually 3 flavors of SCOP, not 2, as described here: https://scop.berkeley.edu/help/ver=2.08#scop2

There's SCOP (version 1) which you could browse both at Berkeley and the UK, but the UK version recently moved to a different website.  There are actually 3 kinds of SCOP identifiers, described here: https://scop.berkeley.edu/help/ver=2.08#stableidentifiers  The sunid is one, but as you noticed in your video, there is also a specific type of family identifier called a sccs, which has the letters and dots.  You can link to SCOP at Berkeley using any of these, as described here:  https://scop.berkeley.edu/help/ver=2.08#linking

The "SCOPe" database is backward-compatible with SCOP 1, and uses the same types of identifiers, and the same URLs.  The only difference is that the database name changed between (SCOP) release 1.75 and (SCOPe) release 2.01.  You can browse both SCOP v1 releases between 1.55 and 1.75 and SCOPe between 2.01 and 2.08 at the same SCOPe website in Berkeley.

The "SCOP2" database is not backward-compatible with SCOP 1, and to make matters more confusing, the maintainers in the UK host SCOP2 at the same URL the used to use for SCOP 1, which breaks all of the old SCOP 1 links.  They did move the old data to a "legacy" website, which serves one older version of SCOP 1 (1.75) - you can see the link on the front of their new site.  I think the old URLs should work if you fix the URL accordingly.

I think your best solution to all of this is to add bioregistry entries for SCOPe (linking to Berkeley), SCOP2 (linking to the UK), and fix the broken SCOP v1 links to explicitly go to SCOP version 1 at Berkeley, e.g., that phage repressor link from your video should go to https://scop.berkeley.edu/sunid=47419&ver=1.75

I'm happy to help if you have more questions!

PR #592, opened 2022-09-28
Update text on health report

cthoyt:
This PR gives more context to the health report and adds coloring in the rows.

See: https://github.com/biopragmatics/bioregistry/blob/update-health-text/docs/health.md


bgyori:
I think it looks good! One note I would add is that if we make references to specific resources/issues, we might have to pay attention to keep this page updated in case their status changes.

cthoyt:
I removed the reference to `bind` because it's sort of complicated. A future revision on this workflow will also add checks for the homepage itself, then maybe we can say something more interesting about that.

PR #600, opened 2022-09-30
Update OMOP

### LINKED ISSUE #599: Update OMOP
### Prefix

omop

### Explanation

- Add upper case prefix as alternative: `OMOP`
- Add https://athena.ohdsi.org/search-terms/terms/ as canonical prefix expansion (nothing better now and everything resolves)
- Optional, I don't care: Add `@prefix omop:       <http://api.ohdsi.org/WebAPI/vocabulary/concept/>.` from prefix.cc (or not, I don't mind)
-  Optional, I don't care: https://faircookbook.elixir-europe.org/content/recipes/applied-examples/ehden-ohdsi.html suggests https://data.ohdsi.org/concept/ as another uri expansion but it does not resolve.
- All OMOP ids are purely numeric, nt sure about the max numbers of digits but OMOP:1 exists. I think the OMOP ID space officially goes to 1,999,999,999 (just short of 2 billion, where the "custom ID space begins).

### Contributor ORCID

0000-0002-7356-1779

cthoyt:
Closes #599 

Thanks a ton @matentzn for finding this information!

@callahantiff you are the OMOP master so could you please take a look at this and suggest any further improvements?

### DIFF
--- a/src/bioregistry/data/bioregistry.json
+++ b/src/bioregistry/data/bioregistry.json
@@ -55784,9 +55784,9 @@
       }
     ],
     "description": "The OMOP Common Data Model allows for the systematic analysis of disparate observational databases. The concept behind this approach is to transform data contained within those databases into a common format (data model) as well as a common representation (terminologies, vocabularies, coding schemes), and then perform systematic analyses using a library of standard analytic routines that have been written based on the common format.",
+    "example": "1",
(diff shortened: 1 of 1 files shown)


matentzn (approved):
Sweet! Thank you!

callahantiff (approved):
Looks good to me! Sorry for the delay in reviewing. 🙌

PR #604, opened 2022-10-02
UniProt proteome

bgyori:
This PR adds a new prefix for UniProt proteomes.


cthoyt (approved):
Awesome, thank you! Looks like the error in the documentation build is unrelated so once that's handled, we can merge this.

PR #631, opened 2022-10-18
Add improved export for `curies`

cthoyt:
This PR does the following:

- [x] implements a more fully-featured generation of a `curies.Converter` class from the Bioregistry
- [x] creates an output for loading the Bioregstry extended prefix map in `curies`
- [x] reimplement reverse prefix map generation
- [x] Rename reverse prefix map (requires updating `curies` package)

CC @caufieldjh

## Blockers

- [x] #632
 
## Follow-up

- create a workflow for generating extended prefix maps for other contexts (notably, the OBO one, but this should be general)


caufieldjh:
Great!

Issue #623, opened 2022-10-11
Campaign to import more of Prefix Commons


cthoyt:
Prefix Commons is a relatively high quality external registry, but it has so much content that we're hesitant to bulk import it. Therefore, we've been thinking about binning its records then importing them in an *ad-hoc* manner, starting with highest quality then moving downwards. Here's a taxonomy of entries so far:

1. resource doesn't exist
2. ready to merge
3. sufficient metadata but some is wrong
4. potential overlaps with bioregistry entries that already exist?

## Phase 1

Start with relatively strict standards where it's unlikely there's lots of curation involved

- check non-trivial
- check URL resolves
- check id pattern
- check manual id

PR #635, opened 2022-10-21
Semi-automatically import more of Prefix Commons

cthoyt:
Related to #623 

This PR imports **68** entries from Prefix Commons with the following qualities:

1. Prefix is 4 or more characters
2. Prefix doesn't already map to something in Bioregistry (i.e., don't add duplicate prefixes)
3. Prefix has full information (name, homepage, pattern, example, URI format)
4. URI formats have to follow "URI prefix" style, which means there's only one `$1` and it comes at the end
5. Homepage and URI format both still work (with a few exceptions)

Manual review was done on most entries. Using `python -m bioregistry.curation.review_pc`. Here were a few tricky bits I ran into:

- several name labels needed to be updated to conform to Bioregistry style guide.
- several (half or more) records constituted "providers" (i.e., they were just reusing a different ID space)
- lots of things had improperly formatted error codes, meaning that redirects happened to some generic warning page rather than giving a proper HTTP 503, 404, etc.


cthoyt:
There are quite a few records in prefix commons that constitute providers. It could be possible to add an additional field in parallel to mappings to keep track of secondary mappings

PR #641, opened 2022-10-25
Enable overriding auto-generated URLs from registries

### LINKED ISSUE #572: Enable overriding default OLS and BioPortal URI construction
As a follow-up to #569, it's the case that several ontologies in the OLS don't use OBO PURLs, which means the automated code for constructing OLS (and BioPortal) URIs is broken.

cthoyt:
Closes #572

There are some overly-generic rules that help generate URIs for various registries. For example, the OLS almost only contained OBO Foundry ontologies at the start of the Bioregistry, so it was a reasonable assumption to make the provider URI for OLS possible to construct in a similar way to OBO Foundry PURLs. This isn't generally the case, so there are a few options:

1. make a more complicated set of rules for each registry where this is the case
2. make a quick and dirty override system, that can be used in a small set of cases.

This PR implements option 2 and uses KISAO as an example.

### DIFF
--- a/src/bioregistry/resolve_identifier.py
+++ b/src/bioregistry/resolve_identifier.py
@@ -468,6 +468,12 @@ def get_iri(
     A custom provider is given, which makes the Bioregistry very extensible
     >>> get_iri("chebi:24867", provider="chebi-img")
     'https://www.ebi.ac.uk/chebi/displayImage.do?defaultImage=true&imageIndex=0&chebiId=24867'
+
+    An override for a metaregistry is built-in to the record
+    >>> get_iri("kisao:0000057", provider="ols")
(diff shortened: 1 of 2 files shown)


src/bioregistry/resolve_identifier.py
```diff
@@ -468,6 +468,12 @@ def get_iri(
     A custom provider is given, which makes the Bioregistry very extensible
     >>> get_iri("chebi:24867", provider="chebi-img")
     'https://www.ebi.ac.uk/chebi/displayImage.do?defaultImage=true&imageIndex=0&chebiId=24867'
+
+    An override for a metaregistry is built-in to the record
+    >>> get_iri("kisao:0000057", provider="ols")
+    'https://www.ebi.ac.uk/ols/ontologies/kisao/terms?iri=http://www.biomodels.net/kisao/KISAO#KISAO_0000057'
```
cthoyt, line 474: should be able to get base IRI from ols, however, bioportal is a completely different situation

PR #642, opened 2022-10-26
Reimplement prefix map generation

cthoyt:
This PR re-implements prefix map generation using the more principled records-based approach, though it required a couple of updates to that to support the same kind of prioritization of prefixes and URI prefixes.


cthoyt:
@matentzn this PR does two things:

1. Better enforces bijectivity of prefix maps
2. Improve the logic used for calculating the priority prefix (this wasn't even possible before) and priority prefix URI

If you take a look at the diff, you'll see that this solves the issue with Geographical Entity Ontology, and the prefix GEO now corresponds to the correct OBO PURL. You can also see a few deletions, which correspond to resources that are duplicates For example, `apid.interactions` is a resource that just reuses UniProt, so it doesn't really make sense to include in a bijective prefix map. This prefix now appears as a synonym for UniProt in extended prefix maps, but no longer in the main outputs.

Simultaneously, I changed the way the schema for contexts works to more explicitly specif that OBO preferred prefixes are the highest priority.

As you're the OBO context czar, please let me know if you're happy with this, or if there are any other considerations for improving the OBO context. One thing I will do in a follow-up PR is make it easier to configure using various forms of Identifiers.org URI prefixes.

matentzn (approved):
I checked the OBO related contexts and they make sense to me, and the changes in general, getting a proper bimap, does make a lot of sense as well!

PR #670, opened 2022-12-06
Add custom Bioregistry deployment documentation

cthoyt:
This PR does the following:

1. Makes it possible to pass custom registry, metaregistry, collections, and contexts 
2. Rewrite the CLI to support more configuration
3. Add tutorial on deploying a mirror of the base Bioregistry
4. Add tutorial on creating custom content and deploying a custom Bioregistry


cthoyt:
CC @ianfore this new tutorial will be the first half of what you'll need to make a custom deployment that also includes custom content. The second half will probably require a discussion to set up.

PR #707, opened 2023-01-12
Fix issues with MIRIAM alignment

cthoyt:
Two issues came up related to recent changes:

1. `miriam:gramene.growthstage` was renamed to `miriam:gro`. This now conflicts with `gro` which is reserved for the Gene Regulation Ontology, so a mismatch was curated and the metaregistry links in the `bioregistry:gramene.growthstage` record was updated
2. `miriam:tair.name` was added, this is equivalent to `bioregistry:araport`. This is the first situation (I think) when Identifiers.org has added a prefix that was a duplicate of something novel in the Bioregistry but didn't have the same prefix. I'm not sure how often this will happen in the future so I'm not sure if this kind of problem needs addressing in the alignment code.


cthoyt:
cc @renatocjn this will be interesting for you

renatocjn:
Well, I can provide some reasoning on the two changes.

First, the gramene.growthstage namespace was renamed in order to make the GRO prefix work. The namespace was non functional anyway. See identifiers-org/identifiers-org.github.io#184 and https://github.com/identifiers-org/identifiers-org.github.io#155. I guess you managed to fix things on your end? I might have to do the same for the oma.hog prefix.

Then, the tair.name prefix was registered by request of the DBCLS team. They argue that TAIR names are also used as IDs. From the homepage of araport prefix that you mention, it seems that it was discontinued and incorporated into TAIR and some other repositories. So saying that it is a duplicate is not exactly correct.

cthoyt:
> Well, I can provide some reasoning on the two changes.
> 
> First, the gramene.growthstage namespace was renamed in order to make the GRO prefix work. The namespace was non functional anyway. See [identifiers-org/identifiers-org.github.io#184](https://github.com/identifiers-org/identifiers-org.github.io/issues/184) and [identifiers-org/identifiers-org.github.io#155](https://github.com/identifiers-org/identifiers-org.github.io/issues/155). I guess you managed to fix things on your end? I might have to do the same for the oma.hog prefix.

That's a bummer, consistency of prefixes is important

> Then, the tair.name prefix was registered by request of the DBCLS team. They argue that TAIR names are also used as IDs. From the homepage of araport prefix that you mention, it seems that it was discontinued and incorporated into TAIR and some other repositories. So saying that it is a duplicate is not exactly correct.

Regardless of what prefix assigned, these represent the same semantic space. Within the scope of the Bioregistry, these are equivalent. The Bioregistry models that multiple prefixes can correspond to the same thing (e.g., `eccode` and `intenz` are the same).

Further, there can be many potential providers of a semantic space. Unfortunately it's very confusing to follow the succession of TAIR's semantic spaces as various resources were shut down/consolidated.

PR #741, opened 2023-02-22
Update license for sphn

deepakunni3:
This is a PR to update the license for the SPHN

@cthoyt Let me know if there is anything else I need to update as part of the PR


cthoyt:
It would be informative to attach an email to your contributor information - is there one that you're okay with being public? Otherwise this is great, another dataset gets its wings!

deepakunni3:
oh yes, of course! Just added the email 👍

cthoyt:
Thanks! Will merge after lunch

deepakunni3:
Thank you for fixing the tests. Was trying locally to see what was the issue.

Much appreciated :)

PR #749, opened 2023-02-28
Add new prefix `emaps`

### LINKED ISSUE #747: Register EMAPS, a variant of EMAPA
We already have an entry for EMAPA: https://bioregistry.io/registry/emapa. EMAPA IDs refer to developmental stage-independent mouse anatomical features, for instance, [EMAPA:35178](https://bioregistry.io/emapa:35178) which resolves to the MGI's Mouse Developmental Anatomy Browser. However, EMAPA has an extension called EMAPS which introduces terms for different developmental stages of this generic EMAPA entry. The IDs are constructed such that they are a concatenation of the EMAPA ID and a Theiler Stage (TS) number. For instance, EMAPA:35178 in TS 23 becomes EMAPS:3517823 which can be resolved at https://www.informatics.jax.org/vocab/gxd/anatomy/EMAPS:3517823.

Worth noting that EMAPS IDs are used in the wild and my motivation for starting the issue is to be able to handle/resolve those IDs.

I started requesting a new prefix at first but then realized that this is a somewhat more nuanced situation that might call for a custom solution. Any thoughts, @cthoyt?

cthoyt:
Closes #747

### DIFF
--- a/src/bioregistry/app/templates/resource.html
+++ b/src/bioregistry/app/templates/resource.html
@@ -442,7 +442,7 @@
                             <a href="{{ url_for("metaregistry_ui.resource", prefix=rel_prefix) }}"
                                class="badge badge-pill badge-light" data-toggle="tooltip"
                                title="{{ manager.get_name(rel_prefix) }}">
-                                {{ manager.get_preferred_prefix(rel_prefix) }}
+                                {{ manager.get_preferred_prefix(rel_prefix) or rel_prefix }}
                             </a>
                         {% endfor %}
                     </dd>


cmungall:
Here are my general recommendations for the EMAPA team:

 * register EMAPS on OBO, and have standard ontology downloads available (we can help with ODK-izing)
 * for the stages themselves (TS1, TS2, ...):
      - it would be wonderful if EMAPA could adopt MmusDv directly (the bgee team would be happy to help cc @fbastian, and ownership could be passed on to Terry)
      -  But I appreciate this may have ramifications in existing users. In this case I would push for the stage terms to have more of a first-class standing in EMAPA. Right now they are not prefixed, which is permitted by oboedit but causes all kinds of issues. Terry could push for a new namespace in OBO for these (analogous to how ZFA also owns ZFS), I don’t love having TS as the namespace, OBO discourages new 2-letter prefixes due to likelihood of conflict

cthoyt:
CC @tfhayamizu 

@cmungall regardless of whether we're able to contact the EMAPS team nor work with them on the OBO community level, we will still be able to put this information into the Bioregistry. I will merge this PR momentarily. Maybe there's a better place to recapitulate this (such as on the OBO foundry issue tracker).

In the mean time, I have removed the `theiler` new prefix request. I also agree TS is a probematic namespace, especially since it conflicts with CALOHA. I would like to solve this by introducing custom prefix remapping in PyOBO when loading ontologies.

PR #754, opened 2023-03-03
Update CALOHA

### LINKED ISSUE #721: Nextprot vs CALOHA confusion
Nextprot maintains a set of tissues with identifiers of the form `TS-\d+` that can be resolved directly in Nextprot as e.g., https://www.nextprot.org/term/TS-0001. This collection of tissues is also referred to as CALOHA. There are a few things I don't fully understand:
- Why does Bioregistry use identifiers like `0012` instead of the native `TS-0012`? The CALOHA OBO file uses the TS- format pretty clearly: https://github.com/calipho-sib/controlled-vocabulary/blob/master/caloha.obo and I am looking at third-party usages of these IDs that are also formatted with the TS- prefix.
- Nextprot provides its own resolver for these terms like the https://www.nextprot.org/term/TS-0001/ example above. So I would like to change the URL over from the "fallback" resolver at https://biopragmatics.github.io/providers/caloha/0001/ which I think is redundant in this case.

cthoyt:
Closes #721 following official updates to the CALOHA OBO file in https://github.com/calipho-sib/controlled-vocabulary/commit/5900c91a5b472e4e5610d4627477403cf9ee5ade


cthoyt:
I'll leave this until end of business day CET in case anyone wants to add a comment

PR #766, opened 2023-03-13
Remove deprecation notice

svempati21:
This text corresponds to deprecated functionality and we're removing it because we don't want people using that code anymore

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,20 +250,7 @@ assert ('neuronames', '268') == parse_iri("http://braininfo.rprc.washington.edu/
 assert ('neuronames', '268') == parse_iri("https://braininfo.rprc.washington.edu/centraldirectory.aspx?ID=268")
 ```
 
-> **Warning**
-> The following functionality is deprecated.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -250,20 +250,7 @@ assert ('neuronames', '268') == parse_iri("http://braininfo.rprc.washington.edu/
...
 
-```python
-from bioregistry import curie_from_iri, parse_iri
-
-prefix_map = {
-   "myprefix": "https://example.org/myprefix/"
-}
-assert ('myprefix', '1234') == parse_iri("https://example.org/myprefix/1234", prefix_map=prefix_map)
-assert 'myprefix:24867' == curie_from_iri("https://example.org/myprefix/1234", prefix_map=prefix_map)
-```
```
cthoyt, line 254: There's an extra blank line you forgot to delete, can you please take that out? Thanks!

svempati21: I fixed this in 0175f6db79e4a3db4019be0d1d7f2600a0eeeb5a

PR #774, opened 2023-03-16
Remove old IRI parser

cthoyt:
This is slow and should be done with a `curies.Converter`

- [x] Remove implementation
- [ ] Check how much unused code this created
- [ ] Bump version


cthoyt:
This was handled in #791

PR #778, opened 2023-03-21
Add CELLxGENE providers for CL and UBERON

cthoyt:
@colinmegill are there other ontologies that can be resolved through this service?

Note this means you can resolve the following with the Bioregistry:

1. https://bioregistry.io/CL:0000062?provider=cellxgene
2. https://bioregistry.io/UBERON:2005080?provider=cellxgene


colinmegill:
Great question — at present, just CL and UBERON, no other ontologies.

PR #789, opened 2023-03-26
Improve federated SPARQL query testing to use docker and add virtuoso test

vemonet:
@cthoyt  I improved the federated query testing to use docker and added virtuoso

The `docker compose up` starts a virtuoso, blazegraph and bioregistry. There is a little catch with virtuoso, you need to run an additional command the first time you start it to enable federated queries

I added an expandable section under installations to explain how to start the stack with docker and enable federated query in virtuoso

I also added a step to start the docker compose stack directly with `docker compose up -d` in the GitHub action workflow for ubuntu (the images we use are not available on ARM windows apparently), all tests are passing: https://github.com/vemonet/bioregistry/actions/runs/4525145166

According to the GitHub action workflow run logs the `test_sparql.py` tests are passing on ubuntu, and skipped as expected on windows:

Screenshot from 2023-03-26 16-43-51

Note that the original query you use with VALUES fails on Virtuoso, so I added a simpler SPARQL query to test the most basic federated querying works, then we can iterate to fix issues related to specific query patterns

The csv parsing was failing due to virtuoso adding double quotes, so I used python built-in `csv.DictReader()` to parse it. And virtuoso was not accepting `application/xml`, so I used `application/sparql-results+xml` which works on blazegraph and virtuoso


cthoyt:
thanks @vemonet ! sorry this fell off the radar before. we should catch up soon

PR #817, opened 2023-04-28
Added Mastodon support

egonw:
Please hold of merging. I cannot get the test framework to work locally with Python 3.9 and depend on the testing here.


cthoyt:
@egonw this looks great in theory! A few follow-ups:

1. Need to add a button to the resource page
2. Should we add some testing to make sure that when this is annotated that it starts with an `@` symbol?

egonw:
@cthoyt, okay, implemented your two requests. I looked at the Wikidata support, but that is beyond my skills. This actually already took more time than I really had this weekend. I hope you can live without Wikidata support for now.

cthoyt:
Wow nice job! I will take care of the rest.

cthoyt:
We'll do wikidata lookup next round of PRs

PR #837, opened 2023-05-13
Add logos to data model

### LINKED ISSUE #832: Allow logo URLs for prefixes
Currently MetaRegistries have URLs. It would also be useful to have URLs for prefixes too, insofar as prefixes are proxies for databases. If this is out of scope then it might be nice to have a formal linkage between prefixes and database/group e.g. in infores so logos can be derived from there. cc @turbomam

cthoyt:
Closes #832

This PR adds logos to the data model for prefixes as suggested by @cmungall. Need to check if any of the other registries besides OBO Foundry provides these

Screenshot 2023-05-21 at 10 02 36


cthoyt:
@cmungall can you please provide a logo annotation that's novel to add as the "first" one?

Issue #840, opened 2023-05-16
Ontology metadata?


udp:
We are using https://raw.githubusercontent.com/biopragmatics/bioregistry/main/exports/registry/registry.json in OLS to resolve CURIEs in ontologies.

However, I think it is unclear whether `uri_format` is the format for a <u>URL</u> from which the resource can be obtained; the format for the <u>ontology IRI</u> of the resource; or both.

For example, for GO, the registry provides:

    "uri_format": "http://amigo.geneontology.org/amigo/term/GO:$1",

which is great if I want to make a GO curie clickable. But if I want to resolve the GO CURIE to the IRI in the ontology, I would instead want to know the official OWL prefix which is `http://purl.obolibrary.org/obo/GO_`.

My current thinking is

1. `uri_format` should actually be `url_format`, because it specifically describes _locations_ at which the resource can be accessed. "uri" is still technically correct but I think you can be more specific here which would make it clear this is should not be used as an ontology IRI.

2. Where a resource is/has an ontology/ontologies, a new section like `ontologies: [...]` should be added providing links to the OWL files, the IRI prefix, and any other ontology-specific metadata which may be pertinent.

I am also not convinced this isn't completely out of scope for the bioregistry :D For example, the OBO foundry config provides this kind of metadata for all OBO ontologies, and OLS provides it for others.

However in this case I'm still not sure how to map between bioregistry<->ontologies in OBO/OLS.

cthoyt:
@udp can we call this done?

cthoyt:
Closing since no response @udp

Issue #698, opened 2023-01-04
Notes from talking with michel


cthoyt:
differences in curation between Bioregistry and Prefix Commons
- prefixes and alternate prefixes being curated
- wasn't clear if the source of the alternate prefix

bioregistry doesn't keep track of 

2. IRIs

start annotating first party IRI pattern

Media type for different templates (HTML, RDF, etc.)

or if content type negotiation is available, you can use the same URI

Propogate alternate, obsolete, erroneous and other IRI formats out of Prefix Commons

Discussion on Dot subspaces
---------------------------
- look at subnamespace in dataset field
- part of collection

**look up omim ontology

Bio2RDF

which of the prefixes - preferred prefix in prefix commons is same used in Bio2RDF - give me list of unique list of all prefixes served one record for

Year last accessible?

Issue #666, opened 2022-12-05
Clarifying the distinction between URI expansion and PURL


matentzn:
It is absolutely critical for data integration to know what _one_ way a CURIE should be expanded back to (in terms of URL).

for example, the `chebi` prefix should _always_ be translated to http://purl.obolibrary.org/obo/CHEBI_ which is what CHEBI.owl is using. 

It seems to me after listening to today's workshop that we should consider distinguishing between:

- PURL
- preferred URI expansion

Maybe these really are different.

balhoff:
I agree—I came here to open an issue but found this existing one. I think Bioregistry should somewhere indicate what the canonical expanded URI form is for an identifier. For example, on the GO page (https://bioregistry.io/registry/go) there are a number of ways listed to turn a GO term CURIE into a URL, but nowhere does it state that the officially endorsed and preferred URI identifier form for a GO term is `http://purl.obolibrary.org/obo/GO_nnnnnnn`.

cthoyt:
So maybe there are two axes we should annotate:

1. Official suggestion from first party
2. This is a PURL

I think it's totally reasonable to keep track of first party annotations. I will be updating the data model in the future also to include a tag if a URI format is appropriate for RDF usage,

Can take responsibility for writing a PR that includes an explanation of what makes something a PURL and what are the curation guidelines for looking at existing URI format strings and determining if they're PURLs? This can go in https://github.com/biopragmatics/bioregistry/blob/main/src/bioregistry/app/templates/meta/summary.html on first try but we could move it later.

matentzn:
I put it on my list to draft a blurp, but I am in a paper jam atm, so not sure how quick I will get to it.

PR #843, opened 2023-05-17
Explicitly annotate URI format for RDF usage

cthoyt:
Related: #840, #666, #698

This PR adds the `rdf_uri_format` field to the Bioregistry's metadata model for each prefix. This can be annotated explicitly, or it pulls from a combination of:

1. OBO Foundry
2. Wikidata (building on #844)
3. Prefix Commons (todo)

For example, this bubbles PURLs assigned by UniProt to the top in this field, whereas the normal URI format string returned corresponds to a URL.

See the updated output in https://github.com/biopragmatics/bioregistry/blob/rdf-uri-format/exports/registry/registry.json


cthoyt:
@udp please have a look at the diff here, we're now explicitly annotating the RDF URI format string in a different slot from the regular URI format string

src/bioregistry/schema/schema.json
```diff
@@ -260,6 +260,11 @@
           "integration_status": "required_for_new",
           "type": "string"
         },
+        "rdf_uri_format": {
+          "title": "RDF URI format string",
+          "description": "The RDF URI format string, which must have at least one ``$1`` in it",
+          "type": "string"
```
egonw, line 266: Does this allow one provider to provide more than one URI pattern/format?

src/bioregistry/resource_manager.py
```diff
@@ -807,6 +807,7 @@ def rasterized_resource(self, resource: Resource) -> Resource:
             example_extras=resource.example_extras,
             example_decoys=resource.example_decoys,
             uri_format=resource.get_uri_format(),
+            rdf_uri_format=resource.get_rdf_uri_format(),
```
egonw, line 810: I like the symmetry with `uri_format`. What are your ideas about URI versus IRI? Or is any IRI also allowed as value anyway?

src/bioregistry/schema/struct.py
```diff
@@ -284,6 +284,10 @@ class Resource(BaseModel):
         description="The URI format string, which must have at least one ``$1`` in it",
         integration_status="required_for_new",
     )
+    rdf_uri_format: Optional[str] = Field(
+        title="RDF URI format string",
+        description="The RDF URI format string, which must have at least one ``$1`` in it",
```
egonw, line 289: Consider linking in the description to the reference specification of the URI (or IRI)

src/bioregistry/schema/schema.json
egonw: But I can also imagine that each URI format is a separate provider. Example: WikiPathways has WPRDF and GPMLRDF and each pathway has a IRI in each one of them, but with a different pattern.

egonw (commented):
Looks like a great start! I left some comments/questions.

Really happy you found time to move this forward (I did not yet). The distinction between URL and URI/IRI is a useful one, I think. I don't think the BridgeDb Open PHACTS work did, but Christian (Uni Manchester) did collect a lot of URLs and IRIs that allows us to link together RDF datasets with the "Query Expander". See `bridgeDB:hasUriPattern` the triples in this file: https://github.com/bridgedb/BridgeDb/blob/master/org.bridgedb.rdf/resources/IdentifiersOrgDataSource.ttl#L4624

src/bioregistry/schema/schema.json
cthoyt: On the first pass, this doesn't allow this. I had considered adding a flag into the "providers" field, where additional ones can be annotated, but I didn't think adding this kind of complexity on the first pass. There's also an argument to be made that the RDF IRIs should be globally unique, but I will let people more invested argue that out ;) 

Can you please give an explicit example for wikipathways that we can come back to later?

src/bioregistry/resource_manager.py
cthoyt: I am not familiar with places where the distinction between URI and IRI makes a difference. If we can identify a place where this matters, we can reinvestigate this question

src/bioregistry/schema/struct.py
cthoyt: - URI format spec: https://www.rfc-editor.org/rfc/rfc3986
- IRI spec: https://www.ietf.org/rfc/rfc3987.txt

src/bioregistry/resource_manager.py
egonw: If it support IRIs, then it can just say IRI. That's my thinking. More precise, more clear. But this can easily be explained in the field description too. Totally fine with that.

src/bioregistry/schema/schema.json
egonw: Yes, will look up two matching IRIs after lunch.

src/bioregistry/schema/struct.py
cthoyt: Done in 0b2d923e

src/bioregistry/resource_manager.py
cthoyt: Done in 0b2d923e

src/bioregistry/schema/schema.json
egonw: - WPRDF: <https://identifiers.org/wikipathways/WP4860>
- GPMLRDF: <http://rdf.wikipathways.org/Pathway/WP4846_r124663>

The latter actually has the revision in the IRI, so that complicates things a lot. So, more than happy to just take the WPRDF URI pattern for now. The context is that the GPMLRDF is the RDF that describes the pathway model (the GPML as RDF), while the WPRDF is the biological process. See https://doi.org/10.1371/journal.pcbi.1004989 for exact details

cthoyt: I don't think this is a very good example to make the case for a given semantic space having multiple RDF URIs - the inclusion of the revision makes it part of a different (albeit, related) semantic space. 

Here's a query over Wikidata for any property that has more than one RDF URI format, it doesn't appear to return any results: https://w.wiki/6j4n

Maybe you have a different example in mind? Like an example where UniProt/SIB has PURLs for something that also has a first-party PURL? @JervenBolleman maybe you know of an instance where this happened

cthoyt:
> This looks great - I think providing both will really help to prevent people from erroneously assuming URLs are RDF IRIs and vice versa. We will definitely use both to inform OLS xrefs.
>
> The only feedback I have would be that maybe the existing `uri_format` should be called `url_format` because it is specifically a resource _locator_, but I understand that a breaking change probably wouldn't be ideal at this point.

As far as I can tell, there's no way to solve the issue looking at one of these things and deciding if it's a IRI, URI, and/or URL. Egon even suggested we rename this field to IRI instead of URI to show it's the most generic possible! So we're going to leave this for now. If you have any ideas how to address this at scale, I'm all ears. Otherwise, suggestions on how to update the docs to make this more approachable for people with different perspectives are also welcome :)

src/bioregistry/schema/schema.json
JervenBolleman: Well the worst case that I can think of is pubmed.  They when using the facebook/open graph protocol, state that the IRI of the pubmed:1 concept is `https://pubmed.ncbi.nlm.nih.gov/1/` in the RDF based OGP.

But there is also `http://rdf.ncbi.nlm.nih/pubmed/1` and there is of course also `http://rdf.ncbi.nlm.nih.gov/pubchem/reference/1`. 

This next to `http://purl.uniprot.org/citation/1`. n2t and identifiers.org. 

Note: the identifier in `http://purl.uniprot.org/citation/X` may be something else than a pubmed identifier.

Note: I once hope to hear from the Pubmed technical staff about the semantics of these different IRIs.

Note: NCBI Taxonomy is similar in it's problem space, but their tech stack is even older and budget to fix anything is minimal.

Note: For UniProt we try our hardest to not use fake uniprot purls for databases that have an RDF representation or thought about the semantics of their IRIs. When they exists, and we don't use them it is most likely because we don't know about them.

egonw: > Well the worst case that I can think of is pubmed. They when using the facebook/open graph protocol, state that the IRI of the pubmed:1 concept is `https://pubmed.ncbi.nlm.nih.gov/1/` in the RDF based OGP.

Yes, another nice example. We have plenty of this issue in Open PHACTS, also for UniProt ;)

But I think we can argue that every pattern is provided by a different provider and "solve" it that way. Not?

JervenBolleman: @egonw depends on what you want to annotate. e.g. provider recommended or just in use?

egonw: I think Bioregistry is all models around providers, so I would say "provider recommended".

Unless you mean one provider recommends one thing and then does another. But that situation is perhaps similar to the above WikiPathways example where this provider (us) use two IRI patterns for different subsets.

cthoyt: Thanks for the nice comments! I think for the purposes of this PR, we can call this done. But if we need to look into this again, this will be the first conversation to link to :)

PR #859, opened 2023-05-20
Cache compiled re pattern

### LINKED ISSUE #858: Slow identifier validation
Resources compile the regular expression pattern for their identifiers each time validation is done. Given that resources are classes, compiled patterns should be cacheable. See these lines with the FIXME comment: https://github.com/biopragmatics/bioregistry/blob/f0af040e8b78da94049b70b095e0fa6cbf2723d7/src/bioregistry/schema/struct.py#LL828C1-L829C35

bgyori:
Fixes #858. My quick benchmark shows that this results in an about 30x speedup when doing a large number of `is_valid_identifier` calls.

### DIFF
--- a/src/bioregistry/schema/struct.py
+++ b/src/bioregistry/schema/struct.py
@@ -270,6 +270,9 @@ class Publication(BaseModel):
 class Resource(BaseModel):
     """Metadata about an ontology, database, or other resource."""
 
+    class Config:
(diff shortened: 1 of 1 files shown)


cthoyt:
@bgyori this is great, had you considered putting a `@lru_cache(1)` decorator on the function instead? Would that have a different effect?

PR #897, opened 2023-06-29
ClinicalTrials.gov: 2023 URI update and ctgov prefix

dhimmel:
refs https://github.com/greenelab/covid19-review/issues/1210

Two changes (one per commit) that should be treated as separate (I can remove either of the changes from this PR based on review).


dhimmel:
> Possible confusion with Connecticut's https://portal.ct.gov/.

ChatGPT answers with ClinicalTrials.gov for "In the context of life science databases, ontologies, and other resources, what does ctgov stand for?"

cthoyt:
@dhimmel thank you for the contribution, the new URI format is great. Can you give me an example where you encountered ctgov as a prefix?

dhimmel:
> Can you give me an example where you encountered ctgov as a prefix?

I don't know if I've seen it as a CURIE prefix anywhere, but I think it is common to use in column names like `ctgov_id`. Also when writing ClinicalTrials.gov, people will often abbreviate it as `CT.gov` or `CT-gov`. Searching via [pubmed](https://pubmed.ncbi.nlm.nih.gov/?term=%22CT.gov%22+OR+%22CT-gov%22+OR+%22ctgov%22), about 25-30 publications do this per year. Admittedly that is low compared to the 20k articles that [mention](https://pubmed.ncbi.nlm.nih.gov/?term=%22ClinicalTrials.gov%22) ClinicalTrials.gov per year.

Partly this addition is motivated by the existing prefixes at https://bioregistry.io/registry/clinicaltrials being somewhat unsatisfactory, where `clinicaltrials` looks out of place as a prefix in the [Clinical Trial Registries collection](https://bioregistry.io/collection/0000012). This change is a potential gateway into a future changing the preferred prefix to `ctgov`, which could be evaluated in the future based on adoption of the `ctgov` prefix.

Also interested in whether anyone knows someone more involved with CT.gov data that could advise? Tagging @micronix in case they have any insights here (I see they are a recent contributor to https://github.com/ctti-clinicaltrials/aact).

dhimmel:
If you're referring to the [giant AACT schema](https://aact.ctti-clinicaltrials.org/static/documentation/aact_schema.png) it looks like it uses both `nct` and `ctgov`.

All good on my end, so merge away. More feedback still welcome, but can always make changes later if anything is actionable.

Issue #910, opened 2023-07-19
Add uri prefix to obo namespace


matentzn:
### Prefix

obo

### Explanation

Previously there was a uri expansion registered in bioregistry, that somehow seems to have disappeared (I also cant see it on the site), see https://github.com/mapping-commons/sssom-py/pull/390/files#diff-fd1c3bdd72413da5a911681114e61f721a38ef02fef95050d17039c16b1ba95cL922

Can we get it back?

```
http://purl.obolibrary.org/obo/
```

### Contributor ORCID

0000-0002-7356-1779

PR #911, opened 2023-07-19
Add URI format to `obo`

cthoyt:
Closes #910

PR #913, opened 2023-07-19
Add uri_format to `obo`

matentzn:
Closes https://github.com/biopragmatics/bioregistry/issues/910

Oops you already did it in #911 -> Not sure if you care about the additional content or just drop it.


src/bioregistry/data/bioregistry.json
```diff
@@ -68117,6 +68117,7 @@
       "fairsharing": "FAIRsharing.847069"
     },
     "name": "Open Biological and Biomedical Ontologies",
+    "pattern": "^[A-Za-z][A-Za-z_0-9]+$",
```
cthoyt, line 68120: so can you give some examples of what is actually valid in this URI space since ther aare actually a lot of things tat get put here

matentzn: Yeah I contemplated that again. I thought this should be a fall-through identifier pattern for everything apart from the gravest violations like leading numbers, punctuation etc (for the case when using bioregistry to validate a curie). But now I am thinking that this may not be enough for example if we want to put ontology purls into CURIE syntax: `obo:mondo/mondo-base.owl`.

Maybe there is no point in documenting a pattern? I always thought it was for validation purposes, and therefore useful, but maybe that's not what it is for?

cthoyt: Yes, the pattern is for validation, but if we put a pattern that's too strict then we will get upset. Bioregistry has a slot for multiple correct example IDs as well as some example incorrect IDs so we can sanity check the pattern. If you make a list of a couple we can test this pattern, otherwise if it's too much work then this one might not be worth bothering

matentzn:
```suggestion
```

I have no meaningful picture of the entire space of possible values - so lets drop it for now and re-introduce it if it becomes needed. Thanks for holding hands!

PR #914, opened 2023-07-19
Add RDF URI format to `GOCHE`

matentzn:


src/bioregistry/data/bioregistry.json
```diff
@@ -40531,7 +40531,8 @@
       "go.chemical",
       "go.chemicals"
     ],
-    "uri_format": "https://biopragmatics.github.io/providers/goche/$1"
+    "uri_format": "http://purl.obolibrary.org/obo/GOCHE_$1",
```
cthoyt, line 40534: it's a mistake to remove this one, since there's no other place to actually resolve these.

matentzn:
```suggestion
    "uri_format": "https://biopragmatics.github.io/providers/goche/$1",
```

PR #916, opened 2023-07-25
Explicitly set MONDO `uri_format` with OBO PURL

dhimmel:
Example resolved URL like
https://monarchinitiative.org/disease/MONDO:0005301

Noticed this when trying to resolve <https://bioregistry.io/MONDO:0002594> and got this [redirection](https://wheregoes.com/trace/20233399988/):

image

Redirects to <http://www.bioregistry.io/0002594> that then says `HTTP 404 - Unknown Prefix`. So confused about why this redirect happens and likely a bug unrelated to the lack of `uri_format`.


cthoyt:
This is an interesting situation. Most of the OBO Foundry ontologies on Wikidata are encoded totally wrong, with bananas everywhere, causing this issue.

All of the OBO Foundry ontologies should get a OBO PURL as their priority URI format, I will look into this

src/bioregistry/data/bioregistry.json
```diff
@@ -61183,6 +61183,7 @@
       }
     ],
     "twitter": "MonarchInit",
+    "uri_format": "https://monarchinitiative.org/disease/$1",
```
dhimmel, line 61186: With `uri_format` would `$1` be replaced with `prefix:identifier` or `identifier`?

> All of the OBO Foundry ontologies should get a OBO PURL as their priority URI format

So should we update so we set `uri_format` to produce a URI like `http://purl.obolibrary.org/obo/MONDO_0004992`? Or will this occur automatically once an upstream change is made?

matentzn: Always use the OBO purl, mondo decides itself through the obo purl system where to redirect

cthoyt: `$1` is always only the local unique identifier.

PR #930, opened 2023-08-29
Fix SSSOM pattern

cthoyt:
The current pattern for SSSOM is incorrect, since it doesn't validate for `superClassOf`

@matentzn does this fix seem alright to you?


cthoyt:
@matentzn other issue, https://w3id.org/sssom/mapping_justification does not resolve to anything :/

PR #948, opened 2023-10-02
Fix typo in README

Adafede:


cthoyt:
Thanks! A long time ago this was correct but I renamed it.

PR #960, opened 2023-10-12
Update PR regex

### LINKED ISSUE #959: Update Protein Ontology regex specification
### Prefix

pr

### Explanation

The given regex in Bioregistry is incomplete. In general there are two types of local identifiers in PR:
- purely digital
- UniProtKB identifier (including isoforms)

The full regex is thus:
^(\d+|[OPQ][0-9][A-Z0-9]{3}[0-9](-\d+)?|[A-NR-Z][0-9]([A-Z][A-Z0-9]{2}[0-9]){1,2}(-\d+)?)$

### Contributor ORCID

0000-0001-5809-9523

cthoyt:
Closes #959

@nataled can you provide a few more example local unique identifiers? I guess that the regex you supplied also includes UniProt isoforms


cthoyt:
@nataled it also appears that `P0DP23` does not match your regex, is this because it only applies to a subset of UniProt?

cthoyt:
It appears the other ones also fail against this regex

nataled:
What's crazy is that I pulled the regex from the OBO Foundry PURL system. In that context they do not fail.

cthoyt:
The test run with Python's builtin `unittest` module's https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRegex. This apparently uses `re.search`. The error message is as follows:

```
    self.assertRegex(example, regex, msg=f"[{prefix}] invalid LUID: {example}")
E   AssertionError: Regex didn't match: '^(?:\\d{9}|[OPQ][0-9][A-Z0-9]{3}0-9?|[A-NR-Z]0-9{1,2}(?:-\\d+)?)$' not found in 'P0DP23' : [pr] invalid LUID: P0DP23
```

I confirmed that if you run the following, nothing is returned

```python
>>> import re
>>> x = re.compile(r'^(?:\\d{9}|[OPQ][0-9][A-Z0-9]{3}0-9?|[A-NR-Z]0-9{1,2}(?:-\\d+)?)$')
>>> x.search('P0DP23')
```

nataled:
I see the problem. Somehow my pasted regex is losing lots of characters, so the one you're testing is an inaccurate regex. I'm pasting the regex into a text file (attached) to avoid this issue. Aligning the two, I see that github didn't much care for the (?: notation, plus did a few other odd removals.

[PR_regex.txt](https://github.com/biopragmatics/bioregistry/files/12896264/PR_regex.txt)

cthoyt:
@nataled looks like the last message did the trick! Thanks!

PR #969, opened 2023-10-30
Add upper- and lowercase prefix synonyms

### LINKED ISSUE #935: Improve EPM export
Add more prefix and punctuation variants to export. E.g., `meddra` should export `MEDDRA` since it's an uppercase variant, even though it's not explicitly annotated

Also, check that preferred prefixes are getting exported properly.

cthoyt:
Closes #935

This PR automatically adds both the upper- and lowercase variants of all prefix synonyms for each record. This makes it much more simple to create comprehensive EPMs (instead of having to refer on programmatic logic for matching)

## Depends on

- [x] #972
- [x] https://github.com/biopragmatics/bioregistry/pull/974
- [x] https://github.com/biopragmatics/bioregistry/pull/975
- [ ] https://github.com/cthoyt/curies/issues/87

### DIFF
--- a/src/bioregistry/data/contexts.json
+++ b/src/bioregistry/data/contexts.json
@@ -2,7 +2,6 @@
   "obo": {
     "blacklist": [
       "icd9",
-      "orphanet",
       "gro"
     ],
     "custom_prefix_map": {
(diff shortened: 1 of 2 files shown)


cthoyt:
The current issue why this isn't working is in the OBO context, the OBO synonyms are prioritized, which gives the Geographical Entity ontology the `GEO` prefix, but doesn't unprioritize the `geo` prefix for Gene Expression Omnibus since this logic is case-sensitive

PR #973, opened 2023-11-02
Update OBO in OWL

cthoyt:
This PR updates the OBO in OWL record (see current record at https://bioregistry.io/oboinowl) in the following ways:

1. Uses pseudo-PURLs for OBO and OWL download that correspond to the URIs
2. Add @kltm as the contact person. There's still some ambiguity as to how this file is maintained, but since the URIs point to the GO repo, I think this is probably a good call for now
3. Adds a loose pattern
4. Adds additional examples

@matentzn @kltm I would love to have one of you contribute a more detailed description

See related recent discussions:

1. https://github.com/geneontology/go-site/issues/2157
2. https://github.com/information-artifact-ontology/ontology-metadata/issues/157


kltm:
I would defer to @cmungall and @balhoff here.

cmungall (approved):

PR #975, opened 2023-11-02
Reimplement CURIE prefix reconciliation for prescriptive contexts

cthoyt:
This now uses the `curies.remap_curie_prefixes` functionality, which is better tested and more flexible.


cthoyt:
cc @matentzn this PR slightly changes the OBO EPM to retain the previous prefix during reconciliation, but this shouldn't have any effect other than to enable better standardization

PR #1003, opened 2023-11-28
Add case variant synonyms to Orphanet and ORDO

cthoyt:
This is an attempt to reconcile Orphanet and ORDO, which have weirdly overlapping synonyms. It's still not clear if these should be merged into a single namespace, or reorganized in a different way


matentzn (commented):
Wait, how will you now deal with the Orphanet identifiers and their uris?

cthoyt:
@matentzn haven't figured this out yet, just seeing to what extent I can add case variant synonyms without breaking anything

src/bioregistry/data/bioregistry.json
```diff
@@ -73514,7 +73512,7 @@
       "version": "4.3",
       "version.iri": "https://www.orphadata.com/data/ontologies/ordo/last_version/ORDO_en_4.3.owl"
     },
-    "pattern": "^C?\\d+$",
```
matentzn, line 73517: With this change, you wont validate any "normal" orpha codes, they don't have that "C"

matentzn: Yes but like I noted, it make more practical sense to push them together instead of having too, else how will you be able to say that both have the same uri_prefix?

matentzn: So far I have not heard a clear argument why the separation is necessary. I have good reasons why it is not _necessary_. I can only think of a single reason, which is being able to redirect xrefs from browsers to separate destinations depending on whether they are properties or diseases. Can you point me to the issue that clearly explains that this is what you are trying to solve?

PR #1005, opened 2023-11-30
Add Bioregistry canonical prefixes as synonyms in OBO export

cthoyt:


cthoyt:
@matentzn do you want to take a quick read through the diff and decide if you think adding these synonyms to the OBO context exports will break anything?

exports/contexts/obo.epm.json
```diff
@@ -189,6 +214,9 @@
     {
         "pattern": "^\\d{8}$",
         "prefix": "APOLLO_SV",
+        "prefix_synonyms": [
+            "apollosv"
```
matentzn, line 218: w/o `_`? I am asking because below the crop ontology ones do have the _

matentzn (commented):
I think this makes sense.. There are potentially issues with this, as we used lower and upper case versions of prefixes to refer to different uri expansions, but I think this practice should probably be abolished.

PR #1027, opened 2024-01-19
Add some synonyms for prefixes

bgyori:
I came across these synonyms being used in ChEBI resource files so I added them here.


cthoyt:
Thanks @bgyori !

PR #1033, opened 2024-01-24
Fix new prefix action code, add missing publications key

bgyori:
This PR adds a mapping from the issue template yaml to an internal data key allowing to process publications submitted as part of the new prefix request form (which was enabled in https://github.com/biopragmatics/bioregistry/pull/1000). This should fix the error in  https://github.com/biopragmatics/bioregistry/actions/runs/7633346253/job/20795375392.


cthoyt:
thanks!

Issue #1032, opened 2024-01-23
Add prefix [hgvs]


miseminger:
### Prefix

hgvs

### Name

Human Genome Variation Society Nomenclature

### Homepage

https://hgvs-nomenclature.org/stable/background/simple/

### Source Code Repository

https://github.com/HGVSnomenclature/hgvs-nomenclature

### Description

The HGVS Nomenclature is an internationally-recognized standard for the description of DNA, RNA and protein sequence variants. It is used to convey variants in clinical reports and to share variants in publications and databases.

The HGVS Nomenclature is administered by the [HGVS Variant Nomenclature Committee (HVNC)](https://hgvs-nomenclature.org/stable/hvnc/) under the auspices of the [Human Genome Organization (HUGO)](https://hugo-int.org/).

### Publications

doi:10.1002/humu.22981 | pubmed:26931183

### Example Local Unique Identifier

NP_003997.1:p.Trp24Cys

### Wikidata Property

P3331

### Contributor Name

Madeline Iseminger

### Contributor GitHub

miseminger

### Contributor ORCiD

 0000-0002-0548-891X

### Contributor Email

miseming@sfu.ca

bgyori:
Thanks, @miseminger! Is there a canonical way to resolve HGVS identifiers? Based on some other registries, I see e.g., http://reg.clinicalgenome.org/allele?hgvs=NP_003997.1:p.Trp24Cys as an endpoint where an HGVS identifier is interpreted. Is there any other that should be prioritized?

miseminger:
Hi @bgyori, thanks for the quick reply!

I haven't been able to find a canonical way to resolve (any HGVS format name), though several different databases use HGVS nomenclature. Your ClinGen link pattern is likely the best one to mention. 

HGVS names each begin with an immutable [reference sequence identifier](https://hgvs-nomenclature.org/stable/background/refseq/), and it allows for a few different reference sequence types and sources, including references from RefSeq, Ensembl, and LRG.

miseminger:
**Update: found one.**

The [Mutalyzer3 API](https://mutalyzer.nl/api/) will take in an HGVS description and give back a JSON file parsing the description; see the `/description_to_model/{description}` tab.

[...]

PR #1037, opened 2024-01-24
Improve redirection

### LINKED ISSUE #1034: Ark identifier example does not redirect via Bioregistry, but does via n2t and identifiers.org
The `ark` prefix (https://bioregistry.io/registry/ark) has an example for which the redirecting works for all but Bioregistry, which goes a 404: https://bioregistry.io/ark:/53355/cl010066723

Preliminary private discussion led to the notion that the `/` at the start (as in `ark:/`) may be the cause.

Relevant code:

https://github.com/biopragmatics/bioregistry/blob/cec184109840de15d1cc0c168845b61a3ae2bb3c/src/bioregistry/app/ui.py#L306-L364

cthoyt:
- Closes #1034 (fixes issue with leading forward slash in ARKs)
- Fixes issue where a colon might appear in the middle of a CURIE that came up in #1032/#1035 for HGVS terms

### DIFF
--- a/src/bioregistry/app/ui.py
+++ b/src/bioregistry/app/ui.py
@@ -303,8 +303,15 @@ def reference(prefix: str, identifier: str):
     )
 
 
+#: this is a hack to make it work when the luid starts with a slash for
+#: ARK, since ARK doesn't actually require a slash. Will break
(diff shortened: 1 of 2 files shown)


egonw:
Thanks!

PR #1042, opened 2024-01-29
Add WB uppercase prefix to bioregistry

matentzn:
The main distributor for gene data on Wormbase gene, the Alliance of genome resources, uses the WB prefix to denote Genes from wormbase, e.g.:

https://www.alliancegenome.org/gene/WB:WBGene00000915


cthoyt:
Thanks!

cthoyt:
I wish we could automate all of this casing (#969) but there are some issues I haven't been able to figure out after several passes with regards to prefix reconciliation for GEO :p

PR #1062, opened 2024-03-25
Add stylization of MeSH prefix to OBO Foundry Community Context

kevinschaper:
Change suggested by @matentzn to get `MESH:` prefix in downstream artifacts


matentzn (approved):
Thank you @kevinschaper! This will percolate within 1 month to all our downstream tooling.

PR #1091, opened 2024-04-15
Add lowercase preferred prefix for semweb resources, update rdf example

### LINKED ISSUE #1090: Update Resource Description Framework
### Prefix

rdf

### Explanation

- the example curie of [rdf:RDF](https://bioregistry.io/reference/rdf:/RDF) is not a thing
- this should be preferred prefix `rdf` (it may seem this is unneccessary but unfortunately bioregistry inherits the incorrect decision to downcase major database prefixes, so it's necessary to annotate all cases where the preferred prefix is truly downcase as such)

### Contributor ORCID

0000-0002-6601-2165

bgyori:
This PR adds the default lower case prefix for each prefix corresponding to a semantic web context resource (see https://bioregistry.io/collection/0000002) as a `preferred_prefix`. It also updates the example identifier for `rdf` to a valid one.

Fixes #1090.


cthoyt:
This all looks good to me. Thanks @bgyori

cthoyt:
Extra note, I ran the following script and noticed that BioLink, LinkML, and NCIT are in LinkML's semantic web list, but not in the Bioregistry's. I'm not sure if it makes sense to add BioLink and LinkML to the semantic web list, but I'm pretty sure NCIT doesn't belong there. Either way, I added preferred prefixes for BioLink and LinkML in 7f899d69 and an additional prefix synonym for `NCIThesaurus` in 96ee766b to cover any future bases.

<details><summary>Code</summary>

```python
import requests
import yaml
import bioregistry

URL = "https://raw.githubusercontent.com/linkml/prefixmaps/main/src/prefixmaps/data/linked_data.curated.yaml"


def main():
    collection = bioregistry.get_collection("0000002").resources

    res = requests.get(URL)
    data = yaml.safe_load(res.text)
    prefixes = set(data["prefixes"])
    for prefix in prefixes:
        xx = bioregistry.normalize_prefix(prefix)
        if xx is None:
            print(f"No prefix in Bioregistry for LinkML prefix: {prefix}")
        elif xx not in collection:
            print(f"Prefix is not in Bioregistry's semantic web collection: {prefix}")


if __name__ == "__main__":
    main()
```

</details>

PR #1096, opened 2024-04-16
Add CLI-based validation of JSON-LD contexts

### LINKED ISSUE #490: Add a CLI method to determine if a JSON context is "bioregistry conformant"
It would be super cool if we could build our own contexts and validate them against bioregistry as part of our CI. I think this would significantly encourage standardisation. I would use this for all projects..

The method should be able to specify a bioregistry managed context by name, and take as an input another context that is validated. Why not use the bioregistry validated context directly? It is likely we will want to keep cached subsets contexts of the huge and growing contexts in bioregistry (specifying maybe 10 prefixes). 

Maybe its a dumb idea but I would use it.

cthoyt:
Closes #490

For example, running the following returns a system exit of 1 because Bioregistry collections indeed are Bioregistry-compliant by construction:

```console
$ bioregistry validate jsonld "https://bioregistry.io/api/collection/0000002?format=context"
```

However, the Prefix Commons vendored Gene Ontology (GO) context is not valid against the Bioregistry. It can be checked with:

```console
$ bioregistry validate jsonld "https://raw.githubusercontent.com/prefixcommons/prefixcommons-py/master/prefixcommons/registry/go_context.jsonld"
BIOMD - nonstandard > Switch to standard prefix: biomodels.db
COG_Function - invalid
WB - nonstandard > Switch to standard prefix: wormbase
FBbt - nonstandard > Switch to standard prefix: fbbt
KEGG_LIGAND - nonstandard > Switch to standard prefix: kegg.ligand
PSO_GIT - invalid
MaizeGDB_stock - invalid
...
```

There are two things that might be the problem. First, the resource might use stylized (i.e. mixed case) prefixes. Therefore, we could try passing ``--use-preferred`` to respect prefix stylization

```console
$ bioregistry validate jsonld --use-preferred "https://raw.githubusercontent.com/prefixcommons/prefixcommons-py/master/prefixcommons/registry/go_context.jsonld"
BIOMD - nonstandard > Switch to preferred prefix: biomodels.db
COG_Function - invalid
WB - nonstandard > Switch to preferred prefix: WormBase
KEGG_LIGAND - nonstandard > Switch to preferred prefix: kegg.ligand
PSO_GIT - invalid
MaizeGDB_stock - invalid
...
```

Second, we could use a pre-defined community context that might have deviations from the vanilla Bioregistry context using the ``--context`` option in combination with one of the contexts' keys (see all contexts here: https://bioregistry.io/context/):

```console
$ bioregistry validate jsonld --context obo "https://raw.githubusercontent.com/prefixcommons/prefixcommons-py/master/prefixcommons/registry/go_context.jsonld"
BIOMD - nonstandard > Switch to preferred prefix: biomodels.db
COG_Function - invalid
WB - nonstandard > Switch to preferred prefix: WormBase
KEGG_LIGAND - invalid
PSO_GIT - invalid
MaizeGDB_stock - invalid
...
```


cthoyt:
@matentzn better late than never, you can use OBO context to validate a JSON with `bioregistry validate jsonld --context obo <url to your JSONLD file>`

PR #1109, opened 2024-05-03
Merge glygen into glytoucan

### LINKED ISSUE #1084: Merge resource GlyGen into GlyTouCan
### Redundant Prefix

glygen

### Target Prefix

glytoucan

### Explanation

The local unique identifier in the GlyGen prefix is actually a GlyTouCan ID, as we use these IDs to identify Glycans in GlyGen.

### Contributor ORCID

0000-0001-8705-181X

bgyori:
This PR merges `glygen` into `glytoucan` while preserving it as a synonym and as a provider for glytoucan IDs. Some issues with this: useful metadata about GlyGen gets lost, for instance its mapping to FairSharing and Integbio, and metadata on a resolver for UniProt IDs on GlyGen.

Fixes #1084


matentzn (commented):
While this seems like the right thing to do, I would not merge this before getting consent from @jeet-vora..

There is also quite a bit of interesting curation here in this record that is lost now - It would be nice if there was a way to set a prefix to "obsolete" to not loose all this information (on the website this would display the metadata with a big warning linking to the replacement namespace). Future music I guess.

bgyori:
Given that the Bioregistry primarily aims to curate identifier resources, and from that perspective, GlyGen really is just a provider, I think these changes make sense. The metadata (i.e., publications, maintainer, etc.) associated with the GlyGen entry is useful but not strictly from a prefix / identifier pattern / resolver URL perspective which again is the main focus here.

bgyori:
Hello @jeet-vora, do you have any thoughts on this change?

bgyori:
Just noticed this is still open, @jeet-vora do you have any comments on this?

jeet-vora:
@bgyori Sorry missed this.

I would keep GlyTouCan and GlyGen as separate entries.
GlyTouCan is a repository for glycans and provides IDs for Them. GlyGen is a knowledgebase that provides information about glycans by using the standard GlyTouCan IDs to represent them. GlyGen also has information about genes and proteins that GlyTouCan does not have. GlyGen does not assign glycan IDs.

Again the above suggestion is based on what I know so far. I my suggestion does not make sense or resolve the issue I might have to dig deep into the communication and the information while I was gone.

bgyori:
I think perhaps our use of the word "merging" is a bit misleading. What happens in this PR is that we keep `glytoucan` as the prefix curared in Bioregistry for a namespace of identifiers while curating `gylgen` as a provider for those identifiers. This, I think, captures what you wrote above.

jeet-vora:
I see. Thanks
Please go ahead as previously planned

PR #1119, opened 2024-05-23
Repaired broken URLs in bioregistry.json

tanayshah2:
This PR updates URLs that were found in a failure report to updated ones that were found.


src/bioregistry/data/bioregistry.json
```diff
@@ -88016,7 +88016,7 @@
       "namespaceEmbeddedInLui": false,
       "pattern": "^PIRSF\\d{6}$",
       "prefix": "pirsf",
-      "uri_format": "https://pir.georgetown.edu/cgi-bin/ipcSF?id=$1"
```
bgyori, line 88019: One important request for change: generally, we shouldn't edit the entries within blocks representing content from other resources mapped to Bioregistry. `n2t` is one example, where in this block, we are simply propagating the URL provided by `n2t`. If we want to override the Bioregistry's prioritized URL, we should curate that as a `uri_format` key at the top level under the given resource's key. This will override any `uri_format` entries from mapped registries like `n2t`.

src/bioregistry/data/bioregistry.json
```diff
@@ -14869,7 +14869,7 @@
     },
     "uri_format": "http://purl.obolibrary.org/obo/CHEBI_$1",
     "wikidata": {
-      "description": "identifier from database and ontology of molecular entities focused on 'small' chemical compounds",
+      "description": "identifier from database and ontology of molecular entities focused on 'small' chemical compounds",
```
bgyori, line 14872: I'm not sure what happened here but there seem to be spurious diffs related to spaces. @tanayshah2 do you know how these were introduced? It's possible that they are changed automatically by an editor you used to edit and save the JSON file. Ideally we should just edit the URLs in question and try to keep everything else unchanged.

src/bioregistry/data/bioregistry.json
```diff
@@ -25396,7 +25396,7 @@
     "publications": [
       {
         "doi": "10.3233/sw-150177",
-        "title": "The Document Components Ontology (DoCO)",
+        "title": "The Document Components Ontology (DoCO)",
```
bgyori, line 25399: A number of strings still have the spurious diffs like this. Could you revert these, or start again with the original bioregistry.json and just apply the specific changes needed to update URLs?

src/bioregistry/data/bioregistry.json
```diff
@@ -56776,7 +56776,8 @@
       "name": "InterPro ID",
       "pattern": "^IPR\\d{6}$",
       "prefix": "P2926",
-      "uri_format": "http://www.ebi.ac.uk/interpro/entry/$1"
+      "uri_format": "https://www.ebi.ac.uk/interpro/entry/InterPro/$1"
+      "uri_format_rdf": "http://www.ebi.ac.uk/interpro/entry/$1"
```
bgyori, line 56780: These changes are still under `wikidata`. We shouldn't change the content within mapped keys like `wikidata`, only outside.

src/bioregistry/data/bioregistry.json
```diff
@@ -60095,8 +60096,9 @@
       "prefix": "lbo"
     },
     "pattern": "^\\d{7}$",
+    "rdf_uri_format": "http://purl.obolibrary.org/obo/LBO_$1",
     "repository": "https://github.com/AnimalGenome/livestock-breed-ontology",
-    "uri_format": "http://purl.obolibrary.org/obo/LBO_$1"
+    "uri_format": "https://bioportal.bioontology.org/ontologies/LBO/?p=$1"
```
bgyori, line 60101: This doesn't seem to actually link to the specific entry, the right URL would be this, I think:

```suggestion
    "uri_format": "https://bioportal.bioontology.org/ontologies/LBO/?p=classes&conceptid=http://purl.obolibrary.org/obo/LBO_$1"
```

src/bioregistry/data/bioregistry.json
```diff
@@ -85403,7 +85405,7 @@
         "year": 2018
       }
     ],
-    "uri_format": "http://purl.obolibrary.org/obo/PECO_$1"
+    "uri_format": "https://browser.planteome.org/amigo/term/PECO:$1"
```
bgyori, line 85408: Here we need to preserve the original PURL as an `rdf_uri_format`

PR #1122, opened 2024-06-03
Bioregistry update visualization script

tanayshah2:
This script visualizes updates in the Bioregistry given two dates, separating them into two bar graphs—one showing updates in main fields, and one showing updates in fields mapped to other registries.


bgyori (approved):
Thanks @tanayshah2, looks great!

PR #1128, opened 2024-06-17
Update omim.ps preferred prefix to OMIMPS in obo context

matentzn:
Most people in the community have moved to using MIM identifiers for OMIM, which we all should. However, some of (like Mondo) have not managed to do that move yet. This change is safe for anyone using the mim strategy.

- [x] needs @anitacaron review


bgyori:
Thanks @matentzn, I was just about to make the change to the order or entries but it looks like you got it. This looks good to me!

matentzn:
As per context protocol best to have the other context owners (@anitacaron) approve first! Thanks!

anitacaron (approved):

PR #1140, opened 2024-07-08
Add prefix genbase

tanayshah2:
This PR adds a new prefix GenBase. GenBase is a genetic sequence database that accepts user submissions (mRNA, genomic DNAs, ncRNA, or small genomes such as organelles, viruses, plasmids, phages from any organism) and integrates data from INSDC.


src/bioregistry/data/bioregistry.json
```diff
@@ -42077,6 +42077,63 @@
     },
     "uri_format": "https://www.ncbi.nlm.nih.gov/nucleotide/$1"
   },
+  "genbase": {
+    "contact": {
+      "email": "genbase@big.ac.cn",
+      "name": "GenBase"
```
bgyori, line 42083: We should use the person listed under fairsharing as the contact since the convention in Bioregistry is to use a specific person rather than a generic contact address.

bgyori (approved):

PR #1142, opened 2024-07-11
title_tfidf.py update

tanayshah2:
This PR contains updates to title_tfidf.py, updating the exports, and implementing functionality to train classifiers and score papers based on abstracts.


bgyori:
A few notes:
- The PR removes two files from the current exports/analyses/tf_idf but it looks like a set of other files still remain in the tf_idf folder (even though the paper_ranking folder is supposed to replace and supersede it. Shouldn't these remaining files also be removed?
- This would be a good time to update the content of outline.png to reflect the current approach
- exports/analyses/README.md should also be updated to reflect the changes
- I am not sure why codecov is erroring, possibly the issue is that there isn't increased test coverage for new code - we may just ignore these

PR #1143, opened 2024-07-15
OMIA: adding new URLs/patterns for OMIA entries

marius-mather:
The aim of this PR is to update OMIA's bioregistry entry to support both types of phenes in OMIA: common entries across species: https://omia.org/phene/omia:001195/, and their appearance in a specific species: https://omia.org/phene/omia:001195-9615/. We have opted for this format, with a single prefix for the two types of phenes, because historically we have not separated them clearly, and have encouraged people to use the same prefix for both.

We have created new URLs on the OMIA site to support these, that should already be working. Where we have existing URLs for entries, the bioregistry URI will redirect to the existing URL.

We also need to create a new prefix for OMIA variant entries (`omia.variant`), but I will do that in a separate PR.


src/bioregistry/data/bioregistry.json
```diff
@@ -79660,7 +79663,7 @@
       "prefix": "omia",
       "uri_format": "http://omia.angis.org.au/retrieve.shtml?pid=$1"
     },
-    "uri_format": "https://omia.org/OMIA$1"
+    "uri_format": "https://omia.org/phene/omia:$1/"
```
matentzn, line 79666: @marius-mather it will create a huge difficulty for us if you add this trailing slash.. can we avoid it? We try to be able to replace prefixes simply, but with a trailing slash we need to add extra logic.. thanks for considering..

matentzn (changes requested):

bgyori:
Thanks @marius-mather, this looks good to me overall, I agree with @matentzn that the trailing slash in the URI format could be removed.

src/bioregistry/data/bioregistry.json
marius-mather: Ah yes, sorry, I think you had mentioned this previously. I think it should be fine but let me do some testing on my end.

marius-mather:
Thanks, I've tested that this will work on our end and removed the trailing slash.

matentzn (approved):
Thanks @marius-mather got accommodating :) lgtm

PR #1145, opened 2024-07-19
Added new prefixes ciliogenics, coi, ferroliganddb, istransbase

tanayshah2:
This PR introduces new prefixes ciliogenics, coi, ferroliganddb, and istransbase.


src/bioregistry/data/bioregistry.json
```diff
@@ -16351,6 +16351,23 @@
...
+      "name": "Mustafa S. Pir"
+    },
+    "contributor": {
+      "email": "shah.tanay2@northeastern.edu",
+      "github": "tanayshah2",
+      "name": "Tanay Shah",
+      "orcid": "0009-0001-1912-5132"
+    },
+    "description": "CilioGenics is an integrated and open source, community friendly database for ciliary genes.",
+    "example": "ENSG00000096872",
```
bgyori, line 16366: This looks like an Ensembl gene ID. Does `ciliogenics` have its own identifier space or is it just resolving Ensembl IDs? If it's the latter, then this could instead be curated as a provider.

src/bioregistry/data/bioregistry.json
```diff
@@ -16351,6 +16351,23 @@
...
+    "homepage": "https://ciliogenics.com/?page=Home",
```
bgyori, line 16367:
```suggestion
    "homepage": "https://ciliogenics.com",
```

bgyori (commented):
One general comment: assuming these new resources were discovered from publications, we should also curate the corresponding papers for each entry.

src/bioregistry/data/bioregistry.json
bgyori: Thanks, I can see that ciliogenics is now listed as a provider for ensembl but I think unless it has its own IDs, we would not list it as a separate entry at all. So if that's the case, we can delete this entry.

PR #1159, opened 2024-08-06
Fix #424

### LINKED ISSUE #424: Fix Plant Ontology URI format
### Prefix

po

### Explanation

I think it should be 
`https://browser.planteome.org/amigo/term/PO:$1` 

instead of

`http://www.plantontology.org/amigo/go.cgi?view=details&query=PO:$1`

The same applies to the homepage. (http://www.planteome.org/ instead of http://www.plantontology.org/)

### Contributor ORCID

0000-0003-0443-9902

Adafede:


docs/_data/health.yaml
```diff
@@ -181308,7 +181308,7 @@ runs:
     failed: false
     prefix: po
     status_code: 200
-    url: http://www.plantontology.org/amigo/go.cgi?view=details&query=PO:0009089
+    url: https://browser.planteome.org/amigo/term/PO:0009089
```
bgyori, line 181311: This report is generated automatically so we don't need to update the URL here

src/bioregistry/data/bioregistry.json
```diff
@@ -89697,7 +89697,7 @@
           "description": "Plant Ontology through Amigo",
           "homepage": "http://www.plantontology.org/",
           "name": "Plant Ontology through Amigo",
-          "uri_format": "http://www.plantontology.org/amigo/go.cgi?view=details&query=PO:$1"
+          "uri_format": "https://browser.planteome.org/amigo/term/PO:$1"
```
bgyori, line 89700: This instance of the URL is ingested from identifiers.org (it is under the `miriam` section) so it isn't meant to be manually updated.

Adafede: Thank you for the prompt feedbak!

Screenshot 2024-08-06 at 14 14 10

How can we change this then?

Adafede: Oh, I see now, sorry my bad, just did a ctrl+F too quickly

src/bioregistry/data/bioregistry.json
```diff
@@ -89781,7 +89781,7 @@
       "synonyms": [
         "poc"
       ],
-      "uri_format": "http://www.plantontology.org/amigo/go.cgi?view=details&search_constraint=terms&depth=0&query=$1"
+      "uri_format": "https://browser.planteome.org/amigo/term/PO:$1"
```
bgyori, line 89784: Similar to the above, this is a URL under `prefixcommons` and not something we would edit here.

bgyori:
Thanks @Adafede! The actual `uri_format` curated directly in the Bioregistry (i.e., not just representing the content of an external registry) is on line 89836 of bioregistry.json:
```
 "uri_format": "http://purl.obolibrary.org/obo/PO_$1" 
```
which is standard for OBO Foundry ontologies and probably not something we want to change. 

In terms of the providers you mention at https://github.com/biopragmatics/bioregistry/pull/1159#discussion_r1705430705, those are currently adopted automatically from an external registry but could be curated directly in Bioregsitry to override any issues there and we can provide more guidance if that's the goal!

Adafede:
Thank you @bgyori! And sorry for not getting it directly... So where should this override be happening? Happy to get more guidance as I think the link at the bottom of the page should be changed

bgyori:
This provider information comes from identifiers.org. The way we can override it is to create a copy of this providers block (marked with ->):
```
 89685     "miriam": {                                                                 
 89686       "deprecated": false,                                                      
 89687       "description": "The Plant Ontology is a structured vocabulary and database resource that links plant anatomy, morphology          and growth and development to plant genomics data.",
 89688       "homepage": "http://bioportal.bioontology.org/ontologies/PO",             
 89689       "id": "00000307",                                                         
 89690       "name": "Plant Ontology",                                                 
 89691       "namespaceEmbeddedInLui": true,                                           
 89692       "pattern": "^PO:\\d+$",                                                   
 89693       "prefix": "po",                                                           
-> 89694       "providers": [                                                            
-> 89695         {                                                                       
-> 89696           "code": "amigo",                                                      
-> 89697           "description": "Plant Ontology through Amigo",                        
-> 89698           "homepage": "http://www.plantontology.org/",                          
-> 89699           "name": "Plant Ontology through Amigo",                               
-> 89700           "uri_format": "http://www.plantontology.org/amigo/go.cgi?view=details&query=PO:$1"
-> 89701         }                                                                       
-> 89702       ],                                                                        
 89703       "sampleId": "0009089",                                                    
 89704       "uri_format": "http://purl.bioontology.org/ontology/PO/PO:$1"             
 89705     },                                                                          

```
directly under the `"po"` entry in bioregistry.json (i.e., one level up from where it currently is), and then modify the `uri_format` value there. I did this on a branch, see here: https://github.com/biopragmatics/bioregistry/compare/main...fix_po but one complication is that because your PR is opened from your main branch I can't push to it directly.

Adafede:
I created a `fix_po` branch but can't open a PR from yours as I am not a collaborator

PR #1163, opened 2024-08-09
Paper Ranking Workflow

tanayshah2:
This PR adds a workflow to automatically create a monthly comment in the Paper Ranking Results issue in GitHub.


.github/workflows/paper_ranking.yml
```diff
@@ -0,0 +1,97 @@
...
+
+    - name: Set up Python
+      uses: actions/setup-python@v3
+      with:
+        python-version: '3.12.2'
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r src/bioregistry/analysis/paper_ranking_requirements.txt
```
bgyori, line 25: Let's leave this as is for now since it's been tested but these could presumably be listed as extra dependencies in the setup configuration instead of in an extra file.

.github/workflows/paper_ranking.yml
```diff
@@ -0,0 +1,97 @@
...
+          const endDate = process.env.END_DATE;
+          const content = fs.readFileSync('exports/analyses/paper_ranking/predictions.tsv', 'utf8');
+          const lines = content.split('\n').slice(1, 21);
+          const rows = lines.map(line => {
+            const [pubmed, title] = line.split('\t');
+            const link = `https://bioregistry.io/pubmed:${pubmed}`;
+            return `| [${pubmed}](${link}) | ${title} |`;
+          });
+          const tableHeader = '| PubMed ID | Title |\n| --- | --- |\n';
+          const commentBody = `This issue contains monthly updates to the ranked list of PubMed papers for curation. The full list can be found [here](https://github.com/${{ github.repository }}/blob/main/exports/analyses/paper_ranking/predictions.tsv). You can view and edit the curated papers file [here](https://github.com/${{ github.repository }}/blob/main/src/bioregistry/curation/curated_papers.csv).\n\n**New entries for ${startDate} to ${endDate}:**\n\n${tableHeader}${rows.join('\n')}`;
```
bgyori, line 79:
```suggestion
          const commentBody = `This issue contains monthly updates to an automatically ranked list of PubMed papers as candidates for curation in the Bioregistry. Papers may be relevant in at least three ways: \n(1) as a new prefix for a resource that can be added to the Bioregistry,\n(2) as a provider for an existing prefix, or\n(3) as a new publication for an existing prefix already in the Bioregistry.\n\nThese curations can happen in separate issues and pull requests. The full list of ranked papers can be found [here](https://github.com/${{ github.repository }}/blob/main/exports/analyses/paper_ranking/predictions.tsv). If you review any of these papers for relevance, you should edit the curated papers file [here](https://github.com/${{ github.repository }}/blob/main/src/bioregistry/curation/curated_papers.csv); these curations are taken into account when retraining the ranking model.\n\n**New entries for ${startDate} to ${endDate}:**\n\n${tableHeader}${rows.join('\n')}`;
```

bgyori:
Oops, I should have done this with a squash merge but forgot to revert that setting.

PR #1164, opened 2024-08-09
Update build and debug update workflow

bgyori:
For several rounds, the automated Update jobs have been failing due to a duplicate URI prefix error due to updated alignment with Wikidata. On this branch, I debugged the issue and added a workaround which resulted in a valid build whose artifacts are pushed to this branch along with a version bump and release.


bgyori:
I'm going to merge this with a merge commit rather than squashing to preserve the automated update commit and the version tag commit for 0.11.11.

PR #1172, opened 2024-08-20
Add new prefix for OMIA variants

marius-mather:
I've added a prefix for genetic variants in OMIA, which are separate from the phenes currently accessible via the `omia` prefix.

This is a new prefix that has not previously been registered with ID services, so I have removed most of the mappings to other services, but let me know if any of these mappings are required.

OMIA variants just have a straightforward numeric ID so the regex pattern for them is simple, and we have added a URL path for resolving them to the OMIA site, e.g. https://omia.org/variant/omia.variant:620/ 

Please let me know if anything further is required for this.


matentzn (approved):
So cool OMIA and you are investing in this @marius-mather.. Really really great.

PR #1147, opened 2024-07-19
Run new prefix requests after debugging

bgyori:
It turns out that if there are any open GitHub issues with the New / Prefix tags and they contain some invalid data, _all_ new prefix issue processing fails before being committed and pull requested. After untagging a problematic issue, the new prefix code runs, the result of which is this PR.

Closes #1146, Closes #1110

PR #1173, opened 2024-08-22
Add reviewer information

lubianat:
Needed for: 
- #1147


lubianat:
Not sure if this is the correct way to add the reviewer info, can you double check it @bgyori ? Thanks!

Issue #1186, opened 2024-09-18
WormBase regex is too restrictive


cmungall:
bioregistry complains about this
https://bioregistry.io/reference/wormbase:/CE00274

but this is a valid wormbase protein ID

The GO registry has info that could be used here but IMO bioregistry could just not bother having a regex for many local IDs, it's hard to keep them up to date

```yaml
- database: WB
  name: WormBase database of nematode biology
  synonyms:
    - WormBase
  rdf_uri_prefix: https://identifiers.org/wb
  generic_urls:
    - http://www.wormbase.org/
  entity_types:
    - type_name: gene
      type_id: SO:0000704
      id_syntax: (CE[0-9]{5})|(WB(Gene|Var|RNAi|Transgene)[0-9]{8})
      url_syntax: http://www.wormbase.org/get?name=[example_id]
      example_id: WB:WBGene00003001
      example_url: http://www.wormbase.org/get?name=WBGene00003001
    - type_name: variation
      type_id: VariO:0001
      id_syntax: (CE[0-9]{5})|(WB(Gene|Var|RNAi|Transgene)[0-9]{8})
      url_syntax: http://www.wormbase.org/get?name=[example_id]
      example_id: WB:WBVar00089843
      example_url: http://www.wormbase.org/get?name=WBVar00089843
    - type_name: RNAi
      type_id: topic:3523
      id_syntax: (CE[0-9]{5})|(WB(Gene|Var|RNAi|Transgene)[0-9]{8})
      url_syntax: http://www.wormbase.org/get?name=[example_id]
      example_id: WB:WBRNAi00110325
      example_url: http://www.wormbase.org/get?name=WBRNAi00110325
    - type_name: Transgene
      type_id: SO:0000902
      id_syntax: (CE[0-9]{5})|(WB(Gene|Var|RNAi|Transgene)[0-9]{8})
      url_syntax: http://www.wormbase.org/get?name=[example_id]
      example_id: WB:WBTransgene00000059
      example_url: http://www.wormbase.org/get?name=WBTransgene00000059
    - type_name: protein
      type_id: PR:000000001
      id_syntax: (CE[0-9]{5})|(WB(Gene|Var|RNAi|Transgene)[0-9]{8})
      url_syntax: http://www.wormbase.org/get?name=[example_id]
      example_id: WB:CE00274
      example_url: http://www.wormbase.org/get?name=CE00274
```

bgyori:
[...]

PR #1187, opened 2024-09-19
Update wormbase pattern

nagutm:
Fixes https://github.com/biopragmatics/bioregistry/issues/1186
@bgyori


src/bioregistry/data/bioregistry.json
```diff
@@ -119164,6 +119164,7 @@
       "homepage": "http://www.wormbase.org",
       "miriam": "wb",
       "name": "Wormbase Gene ID",
+      "pattern": "^(CE[0-9]{5}|WB[A-Z][a-z]+\\d+)$",
```
bgyori, line 119167: @muffymn this is not the right place to put this pattern since it is inside the `wikidata` block. It would have to be one level above this, directly under the `wormbase` entry.

PR #1188, opened 2024-09-30
MolMeDB homepage and repository update

DominikMartinat:


src/bioregistry/data/bioregistry.json
```diff
@@ -70570,7 +70570,7 @@
     "miriam": {
       "deprecated": false,
       "description": "MolMeDB is an open chemistry database about interactions of molecules with membranes. We collect information on how chemicals interact with individual membranes either from experiment or from simulations. ",
-      "homepage": "http://www.upol.cz/en/",
+      "homepage": "https://molmedb.upol.cz",
```
bgyori, line 70573: This edit is within the `miriam` block which we usually don't edit into since it represents what an external source, identifiers.org provides for this resource. The `homepage` entry above, which is directly under the `molmedb` prefix will take precedence over any `homepage` value from external registries. So we can revert the change to this line.

bgyori:
Thanks @DominikMartinat! I just added one comment above if you have time to make that change. If not, I can do it.

src/bioregistry/data/bioregistry.json
DominikMartinat: Thanks. I will have to fix this at identifiers.org. I have tried to change it because the record in the website showed address listed in `miriam` block as MolMeDB homepage.

bgyori (approved):

PR #1189, opened 2024-10-01
Update curated papers list

nagutm:


src/bioregistry/curation/curated_papers.csv
```diff
@@ -1 +1,8 @@
-pmid, relevant, relevancy_type, notes
\ No newline at end of file
+pmid, relevant, relevancy_type, notes
```
bgyori, line 1: I didn't notice this file was originally created but we should remove all the spaces after commas. Further, all text in the notes column should be in double quotes, otherwise the file will not be parseable as valid CSV.

src/bioregistry/data/bioregistry.json
```diff
@@ -118970,6 +118970,7 @@
       "name": "Worm Base",
       "prefix": "426"
     },
+    "pattern": "^(CE[0-9]{5}|WB[A-Z][a-z]+\\d+)$",
```
bgyori, line 118973: This is from a commit that has already been done on another branch so should be removed from this one

nagutm:
Closing Pull Request and creating new one to fix issue of commits from previous branches being visible

PR #1193, opened 2024-10-07
Update curated papers, add new PDB provider (FURNA). 

nagutm:
Updated curated papers list with all papers identified from Aug 9th in https://github.com/biopragmatics/bioregistry/issues/1165.
Curated new provider for PDB Structure called 'furna'. PMID: [39074139](https://bioregistry.io/pubmed:39074139)


src/bioregistry/data/bioregistry.json
```diff
@@ -86670,6 +86679,19 @@
...
+      "email": "nsheffield@virginia.edu",
+      "name": "Nathan Sheffield",
+      "orcid": "0000-0001-5643-4068"
+    },
+    "description": "PEPhub is a database, web interface, and API for sharing, retrieving, and validating sample metadata. PEPhub uses Portable Encapsulated Projects (PEP) biological metadata standard to store, edit, and access PEPs in one place.",
+    "example": "gse185244",
+    "homepage": "https://pephub.databio.org/",
+    "name": "PEPhub",
+    "pattern": "^gse[0-9]{6}$",
+    "uri_format": "https://pephub.databio.org/geo/$1"
```
bgyori, line 86693: Given how this resource is structured, I think we need to curate the `uri_format` as `https://pephub.databio.org/$1`, and allow for an identifier pattern that accepts any `x` or `x/y` combination (and not only gse* which is not generally the pattern for this resource). Then we can curate as example `geo`, and as extra example_extras `geo/gse185244`.

src/bioregistry/curation/curated_papers.csv
```diff
@@ -1 +1,21 @@
...
\ No newline at end of file
+pmid,relevant,relevancy_type,notes
+39104285,1,new_provider,"Provider for UniProt IDs, issue with curation due to multiple URI formats depending on plant species"
+39074139,1,new_provider,"Resolver for PDB IDs"
+39014503,0,no_website,
+39047988,0,irrelevant_other,
+39115390,0,irrelevant_other,
+39095357,0,irrelevant_other,
+39084442,0,not_identifiers_resource,
+38991851,0,new_prefix,"identifiers for sharing, retrieving, and validating sample metadata."
```
bgyori, line 9: Shouldn't this be relevant=1?

src/bioregistry/curation/curated_papers.csv
```diff
@@ -1 +1,21 @@
...
+38991828,0,irrelevant_other,
+39049520,0,not_identifiers_resource,
+39104826,1,existing,"Already present in the bioregistry as a provider for mesh prefix"
+39050757,0,irrelevant_other,
+39064021,0,irrelevant_other,
+39028894,0,not_identifiers_resource,
+39044201,0,not_identifiers,"Potential resource for rare diseases identifiers, but not identifier information"
```
bgyori, line 16: Is the `relevancy_type` supposed to be `not_identifiers_resource` here?

src/bioregistry/curation/curated_papers.csv
nagutm: correct, my mistake

src/bioregistry/curation/curated_papers.csv
nagutm: yes, 'not_identifiers_resource' is what I've been using. I must have accidentally deleted 'resource' when I was editing the file

cthoyt (changes requested):
I don't think that PEPhub makes sense as an entry in the Bioregistry, since it itself has pseudo-curies as identifiers. Instead, it makes sense to make providers for `geo`. It also appears that they prefix GEO entries with other pseudo-prefixes like `bedbase`, which might be some other wrapper service around GEO.

You can see the example where The Gene Curation Coalition also wraps HGNC in https://bioregistry.io/registry/hgnc

Further, this PR is doing a few different things. Please split it up into multiple PRs that are each about a single thing

bgyori:
Thanks @cthoyt, let's discuss options:

> I don't think that PEPhub makes sense as an entry in the Bioregistry, since it itself has pseudo-curies as identifiers. 

GEO is more of an exception than the general rule here in that most entries are not easily recognizable community resources like GEO (e.g., https://pephub.databio.org/xuebingjie1990/excludranges, etc.). From the perspective of "I want to be able to refer to a PEPhub entry via a CURIE and resolve it on the web", I think it makes sense to curate it as a prefix. If it were only providing GEO-type entries, I agree a provider or even a registry would be a better fit. Still, of course this is not a clean cut case either way.

> Further, this PR is doing a few different things. Please split it up into multiple PRs that are each about a single thing

This is something I recommented to @nagutm because the changes here correspond to having reviewed a small batch of papers from the list of predictions for potential relevance. I suggested making PRs with changes on ~10 papers so that the rows added to the curation spreadsheet correspond to changes made in Bioregistry data files in a single PR. I suppose this could be broken up into multiple PRs though where each addition to the curation spreadsheet that results in a specific data update would be its own PR.

nagutm:
Thank you for the input @cthoyt. 

To summarize, the possible courses of action are:

1. Add PEPhub as a prefix. Although majority of the entries are currently related to GEO, if in the future more entries are made that  are independent of that then it makes sense to have a way to resolve other PEPhub entries.
2. Add PEPhub as a provider for GEO. Besides GEO, the remaining identifiers on PEPhub are not very populated and seem to be quite experimental in nature. Example: https://pephub.databio.org/chlescure/new-project?tag=hello. If this remains the case for the near future then curating it as a provider might make sense for now and we could reassess in the future.

For now I have removed the commits relating to PEPhub for this PR and will make a separate PR referencing this discussion until it is resolved.

src/bioregistry/curation/curated_papers.csv
```diff
@@ -1 +1,21 @@
...
+39104826,1,existing,"Already present in the bioregistry as a provider for mesh prefix"
+39050757,0,irrelevant_other,
+39064021,0,irrelevant_other,
+39028894,0,not_identifiers_resource,
+39044201,0,not_identifiers_resource,"Potential resource for rare diseases identifiers, but not identifier information"
+39088253,0,irrelevant_other,
+39119155,0,irrelevant_other,
+39005357,0,irrelevant_other,
+39044130,0,irrelevant_other,
+39010878,0,irrelevant_other,
```
cthoyt, line 21: can you add a column with your ORCID and the date when you did curations?

is there some reference material about the `relevancy_type`? If not, this PR would be a good place to write a markdown file with a full explanation of what this file is, what all of the columns arem how it was generated, and what's in it

nagutm: | **relevancy_type** | **definition** |
|--------------|--------------|
| new_prefix        | A resource for new primary identifiers       |
| new_provider      | A resolver for existing identifiers        |
| new_publication      | A new publication for an existing prefix       |
| not_identifiers_resource      | A database, but not for identifier information      |
| no_website       | Paper suggestive of a new database, but no link to website provided      |
| existing       | An existing entry in the bioregistry        |
| unclear      |  Not clear how to curate in the bioregistry, follow up discussion required     |
| irrelevant_other       | Completely unrelated information       |

This is the working list of terms I am using to label papers. I would appreciate any feedback or corrections on how I can make it more descriptive or concise to ensure that anyone other than myself can use this is a reference in the future.

bgyori: I agree, we can extend the CSV file with an orcid column and a date. The tags are still evolving as we start looking at papers but I like these initial tags. As @cthoyt suggested, pasting the table from your comment above into a README.md file + a bit of text explaining which CSV file it is relevant for, then adding it to version control would be useful.

src/bioregistry/curation/curated_papers.csv
```diff
@@ -1 +1,21 @@
-pmid, relevant, relevancy_type, notes
\ No newline at end of file
+pmid,relevant,relevancy_type,notes,orcid,date
+39104285,1,new_provider,"Provider for UniProt IDs, issue with curation due to multiple URI formats depending on plant species",0009-0009-5240-7463,09/24/2024
```
cthoyt, line 2: A few more suggestions:

1. if something is a new provider, why not explicitly link to the prefix it's a provider for? Same if it's a new prefix - you should link to the new prefix
2. can you add a slot for referencing the PR where the relevant changes is added?
3. Please use standard dates in ISO format YYYY-MM-DD
4. Switch to TSV to increase readability/robustness

src/bioregistry/curation/curated_papers.csv
cthoyt: Then next step is also to add unit tests to ensure the integrity of this file, e.g.

1. checking that required fields are present
2. checking format of data is right (PMID is an integer, ORCID matches specific regex pattern, dates are in YYYY-MM-DD standard format)
3. checking enumeration for relevancy_type is part of a controlled vocabulary

src/bioregistry/data/curated_papers.txt
```diff
@@ -0,0 +1,21 @@
+pmid	relevant	relevancy_type	notes	orcid	date_curated
+39104285	1	new_provider	"Provider for UniProt IDs, issue with curation due to multiple URI formats depending on plant species"	0009-0009-5240-7463	2024-09-24
+39074139	1	new_provider	Resolver for PDB IDs	0009-0009-5240-7463	2024-09-24
+39014503	0	no_website		0009-0009-5240-7463	2024-09-25
+39047988	0	irrelevant_other		0009-0009-5240-7463	2024-09-25
+39115390	0	irrelevant_other		0009-0009-5240-7463	2024-09-26
+39095357	0	irrelevant_other		0009-0009-5240-7463	2024-09-26
+39084442	0	not_identifiers_resource		0009-0009-5240-7463	2024-09-27
```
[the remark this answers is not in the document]
bgyori: Fundamentally, this file was just going to be a simple PMID->relevant curation file to keep track of papers examined and be straightforward input for model training. All the other columns are really just nice to have but not mission critical. I was hoping to keep things simple and things are now going in a direction where we have multiple files in multiple places having to be synchronized and referring to each other to manage this curation. I think the simple 0/1 column is useful in that it summarizes whether this paper should be considered a positive or negative for training purposes without having to have a precise closed-vocabulary statement about what exactly is going on.

src/bioregistry/data/curated_papers.tsv
```diff
@@ -1,7 +1,7 @@
 pmid	relevant	relevancy_type	prefix	notes	pr_added	orcid	date_curated
-39104285	1	new_provider		Provider for UniProt IDs, issue with curation due to multiple URI formats depending on plant species	1193	0009-0009-5240-7463	2024-09-24
-39074139	1	new_provider		Resolver for PDB IDs		0009-0009-5240-7463	2024-09-24
-39014503	0	no_website				0009-0009-5240-7463	2024-09-25
+39104285	1	new_provider	uniprot	Provider for UniProt IDs, issue with curation due to multiple URI formats depending on plant species	1193	0009-0009-5240-7463	2024-09-24
+39074139	1	new_provider	pdb	Resolver for PDB IDs	1193	0009-0009-5240-7463	2024-09-24
```
bgyori, line 3: One issue with having an explicit prefix column is that often there will not be a straightforward mapping there. One paper can be broken up into multiple prefixes, on can be a provider for multiple prefixes. Now we could come up with a more general representation like allow a comma separated list of prefixes to be added. But unless really necessary, all these increase the curation burden and make keeping things in sync more difficult.

src/bioregistry/curation/literature.py
```diff
@@ -0,0 +1,44 @@
...
+    "pr_added",  # links back to the PR where curations were done
+]
+
+
+class CurationRelevance(str, enum.Enum):
+    """An enumeration for curation relevance."""
+
+    #: A resource for new primary identifiers
+    new_prefix = enum.auto()
+    new_provider = enum.auto()
```
cthoyt, line 28: Add docstrings for all of these. You can see the results with

```shell
tox -e docs
open docs/build/html/curation.html
```

tests/test_curated_papers.py
```diff
@@ -0,0 +1,62 @@
...
+
+        self.assertTrue(row["pmid"].isdigit(), msg="PubMed identifier should be an integer")
+        self.assertTrue(row["pr_added"].isdigit(), msg="Pull Request should be an integer")
+
+        # Validate relevant is 0 or 1
+        self.assertIn(row["relevant"], ["0", "1"])
+
+        if row["relevant"] == "1":
+            prefix = row["prefix"]
+            self.assertIsNotNone(prefix, msg="prefix should be set for all relevant entries")
```
bgyori, line 34: I don't think this check is useful... again, my intention was to allow for relevant curations without having to have a perfect and complete assignment to an existing prefix. Let's revert this to not make curation really awkward.

src/bioregistry/curation/literature.py
bgyori: Doesn't this mean we now have to have two separate descriptions of the same concept that have to be in synç? As I commented before, the vocabulary at this stage was meant to be preliminary and will evolve while we explore the initial batch of papers. Once we are confident in a good set that covers all relevant types we come across, we can formalize it as a closed vocabulary and start testing for it.

tests/test_curated_papers.py
cthoyt: Alright, we can simplify this for now. In the future, it would be helpful to include some minimum context in PRs about the motivation and scope of experimental things like this. Alternatively, you can wait to send PRs for experimental projects, or keep them in separate repos to avoid the need for more detailed/stringent review that might be confusing for new contributors.

src/bioregistry/curation/literature.py
bgyori: @nagutm let's add the docstrings here to complete this part

src/bioregistry/data/curated_papers.tsv
```diff
@@ -0,0 +1,21 @@
...
+39104826	1	existing	Already present in the bioregistry as a provider for mesh prefix		0009-0009-5240-7463	2024-10-01
+39050757	0	irrelevant_other			0009-0009-5240-7463	2024-10-01
+39064021	0	irrelevant_other			0009-0009-5240-7463	2024-10-01
+39028894	0	not_identifiers_resource			0009-0009-5240-7463	2024-10-04
+39044201	0	not_identifiers_resource	Potential resource for rare diseases identifiers, but not identifier information		0009-0009-5240-7463	2024-10-04
+39088253	0	irrelevant_other			0009-0009-5240-7463	2024-10-05
+39119155	0	irrelevant_other			0009-0009-5240-7463	2024-10-05
+39005357	0	irrelevant_other			0009-0009-5240-7463	2024-10-05
+39044130	0	irrelevant_other			0009-0009-5240-7463	2024-10-05
+39010878	0	irrelevant_other			0009-0009-5240-7463	2024-10-05
```
bgyori, line 21: @nagutm One final request here: why don't we make the `notes` column last since it has variable length, to make the file a bit easier to read.

tests/test_curated_papers.py
bgyori: Commented out the check for now, we could revisit formally annotating prefixes in the curation file in the future, if there is a use case for it.

bgyori:
The Documentation tests fail with
```
Warning, treated as error:
dot command 'dot' cannot be run (needed for graphviz output), check the graphviz_dot setting
docs-test: exit 2 (15.88 seconds) /home/runner/work/bioregistry/bioregistry/docs> sphinx-build -W -b html -d /home/runner/work/bioregistry/bioregistry/.tox/docs-test/tmp/build/doctrees /home/runner/work/bioregistry/bioregistry/.tox/docs-test/tmp/source /home/runner/work/bioregistry/bioregistry/.tox/docs-test/tmp/build/html pid=2071
  docs-test: FAIL code 2 (24.62=setup[8.73]+cmd[0.00,0.00,15.88] seconds)
  evaluation failed :( (24.70 seconds)
Error: Process completed with exit code 2.
```
but I'm not sure anything related to Graphviz changed in this PR. Possibly a new dependency got pulled in the build which breaks the documentation generation.

cthoyt:
@bgyori potential fix in #1205

tests/test_curated_papers.py
```diff
@@ -0,0 +1,72 @@
...
+                msg="prefix should be standardized for relevant entries",
+            )
+        """
+
+        # Validate relevancy_type is in relevancy_vocab
+        self.assertIn(row["relevancy_type"], self.relevancy_types)
+
+        self.assertRegex(row["orcid"], ORCID_PATTERN)
+
+        # Handle None values values for notes
```
bgyori, line 55:
```suggestion
        # Handle None values for notes
```

cthoyt:
Before finishing this, can you explain what's the concept between the curated_papers.tsv and actually making curations in the bioregistry.json? Are these supposed to be done concurrently, or it was just a exception that one of the curations in bioregistry.json accompanied literature curations here? Is the plan to track which literature curation goes with each additional provider? I don't see an actual addition of the publication to the PDB publication list in bioregistry.json in this file. Would be important to clarify these things and put them in the markdown file in https://github.com/biopragmatics/bioregistry/pull/1195 (and ideally, merge that either together with this PR or in quick succession)

bgyori:
In terms of
> Is the plan to track which literature curation goes with each additional provider? I don't see an actual addition of the publication to the PDB publication list in bioregistry.json in this file. 
I totally agree, every curation going into bioregistry.json should obviously refer to the PMID that it was curated based on. It's just that it wasn't clear where to put a citation that is for a _provider_, not the prefix itself. Would the PMID for Furna go into the publication list for `uniprot`? Or should we extend the provider data model to allow publications tied directly to a provider?

nagutm:
With the addition of the pr_added column in the TSV file I think separating individual curations into separate PRs makes the most sense going forward. It will also help to isolate any ambiguous cases like pephub so that the discussion doesn't get lost in a long thread like this one. 

Does it make sense to create separate PRs for the irrelevant papers added to the TSV file moving forward or group them in with independent curation PRs?

cthoyt:
agreed, let's extend the data model for the provider. See https://github.com/biopragmatics/bioregistry/issues/1207

bgyori:
@nagutm the data model now allows adding publications to providers so please add Furna's publication, then we can wrap this up.

PR #1195, opened 2024-10-09
Add guide for semi-automated curation workflow

nagutm:


cthoyt:
Please include this in #1193 instead of making a new PR

cthoyt:
FYI I updated the README.md in the guides folder with more information on how to write future guides

cthoyt (changes requested):
a few things this needs:

1. add a first paragraph with high-level motivation for why a curator should care about reading this guide / doing this curation
2. add some notes about common mistakes new curators make. for example, mixing up what a database is with what a semantic space is, and how to overcome this confusion
3. give an example curation inside the guide
4. update the link to correspond the the new TSV and update language around it, as well as make sure the columns are updated

PR #1196, opened 2024-10-09
Add prefix: `gmmid`

nagutm:


cthoyt (changes requested):
Please highlight in the description of the new record what is the actual semantic space - this is another instance where there seems to be confusion between curating information about a database, and the Bioregistry's actual goal of curating information about semantic space. 

Here, it appears the URI format (e.g., http://www.gmmid.cn/GeneComposition?params=583) is for a page on allele composition for a gene allele. Therefore, the record should mention something about this in the description.

There are other entity types on this website like http://www.gmmid.cn/GeneComposition2?params=1 for an allele composition for a gene category, and http://www.gmmid.cn/GeneCategory2?params=1272 for a gene category. When you're curating a new resource, do a little exploring to try and identify when there are multiple possible semantic spaces (entity types) that the site has pages for. Then it might be the case that you should curate multiple prefixes about the same database. Then, the style is typically to make a prefix like `gmmid.allele` or `gmmid.category` where there is a dot between the database name and the entity type

nagutm:
Upon revisiting  this curation, I've noticed that the resource appears to be poorly maintained. Many of the webpages aren't resolving as expected which raises questions about the overall quality of the resource. The search function seems buggy and does not make it easy to look up entries for various entities. 

These are 3 examples that @cthoyt noted, but all three resolve to layout pages without any actual information. 
http://www.gmmid.cn/GeneComposition?params=583
http://www.gmmid.cn/GeneComposition2?params=1
http://www.gmmid.cn/GeneCategory2?params=1272

Perhaps, this should be considered a `not_notable` case as discussed here: https://github.com/biopragmatics/bioregistry/pull/1225#issuecomment-2436319452 ?

Appreciate any feedback on this matter

bgyori:
Sure, @nagutm, let's close this then, it can always be revisited later if needed.

PR #1200, opened 2024-10-10
Manual update and resolution of BioDeep issue

cthoyt:


bgyori:
@cthoyt there was another update issue a couple of months ago where I had to run a manual update and also temporarily skip wikidata updates, see the diff in cli.py at: https://github.com/biopragmatics/bioregistry/pull/1164/files#diff-0526e71580c9e632142c58847e63d23d01dfaec7caddd8471c808786b34832cf. The duplicate prefix issue making builds error hasn't been figured out yet but ultimately it would be good to un-skip wikidata again.

PR #1202, opened 2024-10-15
Add aipd as a provider for uniprot

nagutm:
This curation is related to [pubmed:39192607](https://bioregistry.io/pubmed:39192607)

The uri_format for this curation uses an IP address which is unconventional. The identifiers cannot be resolved using the homepage link as far as I can tell. I would appreciate a reviewer's thoughts on whether this could create any potential risks further down the line.


cthoyt (changes requested):
Please add an appropriate entry in the "references" key based on the PubMed you mentioned in the issue

cthoyt:
hi @nagutm, sorry I made a typo earlier and I meant to you to fill out the `publications` list. I updated the testing to give some better feedback and also added a curation guide in https://biopragmatics.github.io/bioregistry/curation/publications that explains the difference

cthoyt (approved):

PR #1208, opened 2024-10-17
Implement object model for provider publications

### LINKED ISSUE #1207: Extend provider data model
Include a list of publications in the `Provider` data model. 

https://github.com/biopragmatics/bioregistry/blob/22f53b4105ab6b5eb0e41ea749582833c6cbcd0c/src/bioregistry/schema/struct.py#L232

This should be easy to incorporate since there's already a model for `Publication`

https://github.com/biopragmatics/bioregistry/blob/22f53b4105ab6b5eb0e41ea749582833c6cbcd0c/src/bioregistry/schema/struct.py#L257

It also needs to get propagated upwards to 

https://github.com/biopragmatics/bioregistry/blob/22f53b4105ab6b5eb0e41ea749582833c6cbcd0c/src/bioregistry/schema/struct.py#L1243

bgyori:
This PR allows publications to be added to providers of a prefix. This is useful since provider-specific publications don't have a natural alternative place to be curated. Fixes #1207.

### DIFF
--- a/src/bioregistry/schema/struct.py
+++ b/src/bioregistry/schema/struct.py
@@ -246,7 +246,7 @@ class Provider(BaseModel):
     )
     publications: Optional[List["Publication"]] = Field(
         default=None,
-        description="A list of publications about the provider",
+        description="A list of publications about the provider. See the `indra` provider for `hgnc` for an example.",
     )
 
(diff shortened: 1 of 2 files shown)


cthoyt:
thanks @bgyori

PR #1215, opened 2024-10-20
Add prefix: `scancerrna`

nagutm:


cthoyt:
Why did you choose `scancer` instead of `scancerna` as the prefix?

src/bioregistry/data/bioregistry.json
```diff
@@ -100737,6 +100737,33 @@
...
+    "pattern": "^\\d{1,4}$",
+    "publications": [
+      {
+        "doi": "10.1093/gpbjnl/qzae023",
+        "pubmed": "39341795",
+        "title": "SCancerRNA: Expression at the Single-cell Level and Interaction Resource of Non-coding RNA Biomarkers for Cancers",
+        "year": 2024
+      }
+    ],
+    "uri_format": "http://www.scancerrna.com/BioMarker/toDetail?id=$1"
```
cthoyt, line 100765: the curation of the prefix should be about the entity type that the prefix describes. Here, this is corresponding to a biomarker in this database. The description of a bioregistry record should describe the semantic space, not the database.

src/bioregistry/data/bioregistry.json
```diff
@@ -100737,6 +100737,33 @@
...
+      "email": "m.naguthana@hotmail.com",
+      "github": "nagutm",
+      "name": "Mufaddal Naguthanawala",
+      "orcid": "0009-0009-5240-7463"
+    },
+    "description": "SCancerRNA database contains experimentally supported data on long ncRNA, microRNA, PIWI-interacting RNA, small nucleolar RNA, and circular RNA biomarkers, as well as data on their differential expression at the cellular level.",
+    "example": "9530",
+    "homepage": "http://www.scancerrna.com/BioMarker/",
+    "name": "SCancerRNA",
+    "pattern": "^\\d{1,4}$",
```
bgyori, line 100756: Not sure if we should set 4 as an upper limit here. They don't use 0 padding with fixed length for IDs and they increase incrementally so they could eventually go beyond 4 digits. `^\\d+$` might be more robust here.

src/bioregistry/data/bioregistry.json
bgyori, line 100765:
```suggestion
    "uri_format": "http://www.scancerrna.com/toDetail?id=$1"
```

bgyori: It turns out the `BioMarker` suffix is redundant, all of this resource is about biomarkers and the URLs work without the extra BioMarker suffix.

src/bioregistry/data/bioregistry.json
nagutm: Agreed, updated regex.

src/bioregistry/data/bioregistry.json
nagutm: updated the description to better describe the semantic space.

bgyori:
I checked again and noticed that the prefix should ideally be `scancerrna` (with 2 r-s) to match the resource name so changed that. Everything else looks good to me so I added reviewer info.

PR #1217, opened 2024-10-22
Add publication for `sgd` and `sgd.pathways`

nagutm:
This PR adds publication [pubmed:39345624](https://bioregistry.io/pubmed:39345624) to `sgd`.

Unsure whether this should also be included under the publications for `sgd.pathways`.


cthoyt:
If the publication is about a database then it makes sense to a did the publication for all prefixes that are part of that database. Please open a separate issue to discuss canonicalization of the publication list

cthoyt (approved):

PR #1225, opened 2024-10-24
Add new provider for `hgnc.symbol`: `cbgda`

nagutm:
This resource does create unique identifiers in the format CB1, CB104, etc. as seen in the image below. However, the unique identifiers for each entity are not resolvable through this format of identifiers but rather the name of the gene for `cbgda.gene` and the name of the disease for `cbgda.disease`. 

image


src/bioregistry/data/bioregistry.json
```diff
@@ -12843,6 +12843,62 @@
...
+      "name": "Jinku Bao"
+    },
+    "contributor": {
+      "email": "m.naguthana@hotmail.com",
+      "github": "nagutm",
+      "name": "Mufaddal Naguthanawala",
+      "orcid": "0009-0009-5240-7463"
+    },
+    "description": "This collection represents key host genes identified through genome-wide CRISPR screening as being involved in disease pathogenesis. This space includes manually curated gene data, such as expression patterns, gene variants, and their associations with diseases.",
+    "example": "KAT7",
```
cthoyt, line 12886: is this not just an HGNC gene symbol?

src/bioregistry/data/bioregistry.json
```diff
@@ -12843,6 +12843,62 @@
...
+      "name": "Jinku Bao"
+    },
+    "contributor": {
+      "email": "m.naguthana@hotmail.com",
+      "github": "nagutm",
+      "name": "Mufaddal Naguthanawala",
+      "orcid": "0009-0009-5240-7463"
+    },
+    "description": "This collection represents diseases linked to host genes identified via genome-wide CRISPR screens. It includes detailed disease classifications, gene-disease associations, and integrated data on genetic factors contributing to disease development.",
+    "example": "Glioblastoma",
```
cthoyt, line 12858: We should have a discussion about if this qualifies as a semantic space or not. Combine with the fact that it's not a notable (or, from the UI design, high quality) resource, maybe we should come up with some criteria for skipping this kind of resource. This is similar to the fact that we don't import all of the Bioportal ontologies because a lot of them are junk

cthoyt (changes requested):
1. Have discussion about updating policy on criteria for notability + correctness
2. Update curation guidelines for semi-automated literature curation accordingly
3. Check if the gene namespace is just a provider for hgnc.symbol

bgyori:
I agree with the general idea of using discretion to determine notability. We should probably come up with a curation tag in  the paper curation tsv which expresses something like "relevant but not notable" (this would be a positive training sample for machine learning purposes but something that ended up not being added to the Bioregistry). 

In this case, this is a resource published in Database (Oxford) with a working website so I wouldn't dismiss it apriori as not notable, and the content seems to be pretty useful. Overall, curating it simply as a provider for `hgnc.symbol` would be appropriate.

nagutm:
I think that `not_notable` should work as a tag to describe these types of resources for the future. If you agree, I can move forward with updating the relevancy vocabulary in the necessary files. 
@bgyori

src/bioregistry/data/bioregistry.json
```diff
@@ -49649,6 +49649,22 @@
       "uri_format": "https://www.genenames.org/cgi-bin/gene_symbol_report?match=$1"
     },
     "providers": [
+      {
+        "code": "cbgda",
+        "description": "Key host genes involved in the pathogenesis of different diseases identified through genome-wide CRISPR screens",
```
bgyori, line 49654:
```suggestion
        "description": "A data base of key host genes involved in the pathogenesis of different diseases identified through genome-wide CRISPR screens",
```

src/bioregistry/data/bioregistry.json
```diff
@@ -49649,6 +49649,22 @@
...
+        "homepage": "http://cbgda.zhounan.org/main",
+        "name": "CRISPR Based Gene-Disease Assosciations",
```
bgyori, line 49656:
```suggestion
        "name": "CRISPR Based Gene-Disease Associations",
```

bgyori:
> I think that `not_notable` should work as a tag to describe these types of resources for the future. If you agree, I can move forward with updating the relevancy vocabulary in the necessary files. @bgyori

Yes, we can define a tag like that on a separate PR and use it in the future whenever appropriate.

PR #1227, opened 2024-10-24
Update semi-automated literature curation checks

cthoyt:
1. Require every row to reference the pull request in which the row was added
2. Standardize prefix used to refer to PubMed
3. Enforce sorting (and add reformatting command to `bioregistry lint`)


bgyori:
I'm confused at this point about the role of the `pr_added` column. I had though it referred to the PR where any prefix/provider/publications corresponding to the curated row was added to bioregistry.json. This change makes it seem like the `pr_added` tracks what PR a line in the curation spreadsheet was added. But that doesn't really make sense since git does this naturally... Could you clarify why this change is needed?

cthoyt:
This column describes in what pull request the row in `curated_papers.tsv` was added. I've tried to update the unit test to give this feedback more clearly for when it's missing.

This is important to make it possible to quickly go from the curation TSV to the PR, where the context for the curation itself happened, and also enable discussion (even after the PR has been merged). It's also possible to use `git blame` to look at a given row and eventually reconstruct where it came from, but this isn't convenient in most situations.

Adding the PR where something was curated is becoming common practice in the OBO Foundry. For example, on this OMO term, you can link back to the place where the curation was added: https://www.ebi.ac.uk/ols4/ontologies/omo/properties/http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FOMO_0003003?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FOMO_0003003.

Issue #1222, opened 2024-10-22
General policy for curating shared information for subdivided prefixes


bgyori:
We have run into this issue in a couple of different settings now so we might want to discuss a general approach to the problem of metadata curated across semantic spaces for a given database. 

Example: `pubchem`.
- When a resource like `pubchem` is subdivided into multiple semantic spaces including `pubchem.substance` or `pubchem.compound`, these prefixes each contain some combination of information that is
   - specific to the prefix (e.g., `uri_format`, `example`, `pattern`) and others that are almost always
   - generic for the entire database (e.g., `publications`, `homepage`, `contact`, etc.).
  
- There is also the question of mappings:
  - some mappings are semantic space-specific (e.g., `pubchem.compound`'s mapping to `miriam`'s `pubchem.compound` or to `n2t`'s `pubchem.compound`) while other mappings are 
  - generic for the entire database e.g., `pubchem.compound`'s mapping to `fairsharing`'s `FAIRsharing.qt3w7z`)

Currently, the data that is generic for the entire database is not propagated in a predictable way into semantic space-specific prefixes. The fact that some data is surfaced via prioritized mappings makes the situation more complicated. For example, we have
- [pubchem.compound](https://bioregistry.io/registry/pubchem.compound) listed with 8 publications in total
- [pubchem.substance](https://bioregistry.io/registry/pubchem.substance) with only 1 of those 8 publications
- To further complicate things, 3 of the 8 papers attached to `pubchem.compound` are specifically about PubChem's BioAssay subset for which there is a dedicated prefix at [pubchem.bioassay](https://bioregistry.io/registry/pubchem.bioassay) but that record again only refers to the same 1 publication that `pubchem.substance` does.

So the main question of this issue is: should this type of data be standardized based on its shared vs subspace-specific status? If so, what should be the general policy for this?

This is relevant for e.g., #1214 and #1204 and many existing records.

cthoyt:
see the addition to curation guidelines added https://github.com/biopragmatics/bioregistry/pull/1217

cthoyt:
See also the `part_of` field, described at https://biopragmatics.github.io/bioregistry/datamodel/#part-of where the instances you described plus additional ones have been curated. You can get a quick overview on https://bioregistry.io/highlights/relations

more generally I've considered how to share information between related resources (not just publications, but also homepage, contact person, etc.). Open to suggestions but having a curation script that copies publications is a quick and dirty solution

PR #1232, opened 2024-10-25
Add part_of for discovered families of prefixes

bgyori:
This PR adds `part_of` relationships between prefixes and their assumed parents where missing. The PR stems from the discussion in https://github.com/biopragmatics/bioregistry/issues/1222. I did the following:
- Group prefixes by their root (e.g., `aop` for `aop.events`, `aop.stressor`, etc.) into families. Spot checking these relations show that these appear to be genuine families of prefixes, though there could be exceptions.
- Check which elements of a family already have a `part_of` relationship to the root
- Under the condition that a member doesn't have a `part_of` relationship to a root and the root is also a Bioregistry prefix, add a `part_of` relationship from the given prefix to the root prefix.

Questions to consider:
- We could extend this to add `part_of` relations to parents that aren't Bioregistry entries. An example that already exists is `pubchem`.
- Is the code that made these changes useful at all to be in version control? It might just be a one-off so I am not immediately adding it anywhere.
- Are there any potential concerns where a `part_of` added here isn't actually correct, i.e., where two prefixes with `.`-s in them share a root but shouldn't actually have a `part_of` relationship to the same root?
- As https://biopragmatics.github.io/bioregistry/datamodel/#part-of points out, there are family relationships other than the ones following the `.` separation pattern which aren't picked up here - these might require manual curation.


bgyori:
The test failure appears to have to due with a conflict between `has_canonical` and `part_of` relationships. In particular, `insdc.run` is annotated as `has_canonical` for `ena.embl` but when adding `insdc.run`-`part_of`- `insdc`, it results in `insdc.run` not being surfaced in the top-level of the registry anymore which is what the test appears to expect.

cthoyt:
I removed a few conflicting parts and added a test to make sure the `part of` relationship can't clobber the `canonical for` one. We can come back to the corner case for INSDC in a follow-up. Thanks for the PR, sorry for the long time to get back to it

Issue #1204, opened 2024-10-17
Extend iedb to multiple semantic spaces


bgyori:
### Prefix

iedb

### Explanation

The [IEDB database](https://www.iedb.org/) is already registered as [`iedb`](https://bioregistry.io/registry/iedb). However, IEDB actually consists of identifiers for multiple semantic spaces:
- `reference`: these are internal IDs for sources of evidence like a publication, example: https://www.iedb.org/reference/1038233. This is what is currently assumed as the basis for the URI format in the Bioregistry.
- `assay`: a specific experimental assay serving as evidence for an entry. Example: https://www.iedb.org/assay/1691753
- `antigen`: provides information on a given antigen. Effectively operates as a provider for external IDs. Examples: https://www.iedb.org/antigen/UNIPROT:P62315 and https://www.iedb.org/antigen/ChEBI:143712
- `epitope`: part of a protein that is studied as an epitope. Example: https://www.iedb.org/epitope/117395
- `receptor`: receptors studied as a group. Example: https://www.iedb.org/receptor/47
- `mhc`: MHC restrictions of cells. Example: https://www.iedb.org/mhc/86

and currently the latter five are not appropriately registered. We could consider breaking up `iedb` into `iedb.reference`, `iedb.assay`, etc. without a parent `iedb` entry. This would follow the template of e.g., the `pubchem.*` entries in the Bioregistry.

### Contributor ORCID

0000-0001-9439-5346

bgyori:
A few issues related to this:
- Mappings for the current `iedb` entry:
  - `miriam`: this maps to what would become `iedb.reference`
  - `prefixcommons`: this maps to what would become `iedb.epitope`
  - `fairsharing`, `integbio`, `re3data`: these entries are all generic to all of IEDB so not clear what to do with them if we break up `iedb` into multiple semantic-space-specific entries. a) drop them, b) map them to one arbitrary new `iedb` entry or c)  map them redundantly to all of the new `iedb.*` entries?
- Publications
  - There are 4 publications curated directly under `iedb` (plus additional ones listed under the mapped entries). These publications are all generic to all of IEDB so again not clear what happens if `iedb` is split up. a) do we add them to one arbitrary new iedb entry or b) redundantly add them to all entries? Looking at `pubchem.*` as an example in this regard I didn't find any patterns to follow, publication assignment to each entry there looks pretty arbitrary.

bgyori:
Done in #1238

PR #1238, opened 2024-10-31
Extend `iedb` prefix to multiple semantic subspaces

nagutm:
This PR addresses https://github.com/biopragmatics/bioregistry/issues/1204 by extending the `iedb` prefix into several specific subprefixes to better capture distinct semantic spaces. The subprefixes introduced in this PR are: `iedb.reference`, `iedb.assay`, `iedb.epitope`, `iedb.receptor`, and `iedb.mhc`.

Two main approaches were evaluated for organizing these semantic subspaces:
1. **Retain `iedb` as a parent prefix**: Under this approach, iedb serves as a source of general information, while each subprefix is curated for specific semantic spaces. This structure preserves generic mappings and publication data under the parent iedb prefix, reducing duplication and creating a single reference point for general resources. Each subprefix has a `uri_format` and `pattern` associated with it while the parent prefix does not.

2. **Eliminate the parent prefix and duplicate information**: Alternatively, we could remove the iedb parent prefix entirely and duplicate general information, such as mappings and publications, across each subprefix. While this approach ensures each subprefix is independently comprehensive, it introduces a high amount of redundancy.

This PR implements the first approach, maintaining `iedb` as a parent prefix. The following chages were made:
- General mappings such as `fairsharing`, `integbio`, `re3data` are retained under the parent, while subprefix-specific mappings (`miriam`, `prefixcommons`)  are assigned to its related sematic subprefix.
- Publications that apply broadly across IEDB are preserved under the parent prefix. Any future publications relevant only to specific semantic spaces could be added to the appropriate subprefix as needed.
- `iedb.antigen` is curated separately as a provider for UniProt and ChEBI

One drawback of this approach is that it is not clear what sematic space the parent perfix represents, however as the bioregistry continues to grow larger in the future, I felt that an approach that reduces overall redundancy would be more useful.


nagutm:
Per https://github.com/biopragmatics/bioregistry/pull/1217#issuecomment-2428413739, I duplicated the publications entry in all subspaces

src/bioregistry/data/bioregistry.json
```diff
@@ -54703,6 +54683,257 @@
...
+      },
+      {
+        "pubmed": "16312048",
+        "title": "The immune epitope database and analysis resource: from vision to blueprint",
+        "year": 2004
+      }
+    ],
+    "uri_format": "https://www.iedb.org/assay/$1"
+  },
+  "iedb.epitope": {
```
cthoyt, line 54730: this record needs to retain the miriam link

bgyori: The identifiers.org record appears to be specific to `iedb.reference`, could you elaborate why `iedb.epitope` should retain that link?

cthoyt: might be that the diff is too hard to read, it looked like the URI pattern from the MIRIAM record matched the `iedb.epitope` entry. Will take another look later

bgyori:
The test failure is due to the fact that according to strategy 1 outlined by @nagutm above, the parent `iedb` entry is preserved but it doesn't come with an `example` which is required per the test. This all relates to the more general issue at #1222, whether parent entries that don't correspond to a specific semantic space (which would have a specific pattern, uri_format, example, etc.) should be a thing, or in a situation like this with `iedb`, the parent entry should be completely dissolved.

nagutm:
After discussing with @bgyori  we decided that strategy 2 would be better suited for the following reasons:
- Clear semantic spaces: Each subprefix clearly defines a sematic space without any ambiguity as to what the parent prefix represents
- Alignment with current testing standards: This method prevents us from having to modify current established testing standards that require certain fields to be present for a new prefix such as `example`

PR #1244, opened 2024-11-01
Rewrite `dpvocab` prefix content

nagutm:
This PR fixes https://github.com/biopragmatics/bioregistry/pull/1240#issuecomment-2452171994 which was caused by a buggy github-actions that forcefully overwrote the content of the `dpvocab` prefix with another prefix.


bgyori:
Note: this is a merge into the open PR not main.

PR #1251, opened 2024-11-04
Update arxiv wikidata URLs to HTTPS

bact:
To match current info in Wikidata
https://www.wikidata.org/wiki/Q118398

- `homepage` -> Wikidata's "official website"
  - `https://arxiv.org`
- `uri_format` -> Wikidata's "interwiki prefix at Wikimedia/arxiv/formatter URL"
  - `https://arxiv.org/abs/$1`


cthoyt:
Hi @bact - this PR might not be necessary since the Wikidata fiends are automatically pulled in on an automated script (and will get overwritten). I will try and run that script locally and see what happens

bact:
I think it will be better to try editing at the Wikidata first. I will try updating that.


If there's an automated way to pull data from Wikidata, we can close this PR without merging. Thank you.

cthoyt:
Yes, there is! You can run `python -m bioregisry.external.wikidata.__init__`. However, there's a minor issue I haven't addressed with semi-automated data integrity checks (there are two different records with the same RDF URI, which causes an issue)

bact:
Follow up at the Wikidata page:
https://www.wikidata.org/wiki/Property_talk:P818#Updating_%22source%22,_current_one_is_broken

Issue #1247, opened 2024-11-03
Register FEMA and JECFA numbers


bgyori:
Similar to #1246, encountered FEMA and JECFA numbers as cross-references for compounds.

FEMA: https://www.femaflavor.org/, e.g., 4667, unfortunately, the URLs don't use the ID but rather the name, e.g., https://www.femaflavor.org/flavor-library/24-ethyl-426-dimethyldihydro-135-dithiazine-mixture-isomers.

JECFA: https://apps.who.int/food-additives-contaminants-jecfa-database/, e.g, 2116, again, unfortunately, the URLs don't use the JECFA ID (but some other number that is completely different): https://apps.who.int/food-additives-contaminants-jecfa-database/Home/Chemical/6163. This actually appears in WikiData as P4852 (see external/wikidata/processed.json) but I believe the WikiData record confounds the numbers used in the URL with what is actually called a JECFA ID.

nagutm:
I can look into this

nagutm:
JECFA:
- Although the URL uses a different number than the JECFA ID, not all entries in this resource contain a specific JECFA ID [example 1](https://apps.who.int/food-additives-contaminants-jecfa-database/Home/Chemical/1370), [example 2](https://apps.who.int/food-additives-contaminants-jecfa-database/Home/Chemical/62) that is distinct from the one in the URL.
- Given this, I think that we can curate this in the bioregistry with the number in the URL as the main identifier.

FEMA:
- Although this resource provides references to the FEMA, JECFA and CAS numbers, none of these identifiers are resolvable through the URL. 
- Given that this is a notable resource, we could curate it with the name of the chemical as the main identifier and possibly add the FEMA number in the `example_extras` field.

bgyori:
Thanks for looking into this @nagutm. I would approach this not from the perspective of the URI patterns (which seem to use arbitrary IDs that are not defined anywhere) but from the perspective of the actual FEMA and JECFA IDs (which are defined and appear in cross-references and therefore necessitate a prefix to be representable). So let's curate new prefixes for the FEMA/JECFA IDs and for the time being assume that there is no URI pattern to resolve these - unless of course we find a provider somewhere.

bgyori:
After a bit more digging I actually found that WikiData has two different properties for JECFA, including this: https://www.wikidata.org/wiki/Property:P9557 ("JECFA number") which is the one we should be mapping to instead of P4852.

bgyori:
Done in #1254 and #1255.

PR #1255, opened 2024-11-04
Add prefix: `jecfa`

nagutm:
This pull request addresses part of https://github.com/biopragmatics/bioregistry/issues/1247 by introducing a new prefix `jecfa`.

Since the URL for individual entries does not use the JECFA Number, but rather an unrelated number, it is curated without a `uri_format` field.


src/bioregistry/data/bioregistry.json
```diff
@@ -59265,6 +59265,37 @@
       "uri_format": "http://nikkajiweb.jst.go.jp/nikkaji_web/pages/top_e.jsp?CONTENT=syosai&SN=$1"
     }
   },
+  "jecfa": {
+    "contact": {
+      "email": "jecfa@who.int",
+      "name": "JECFA Secretariat at WHO"
```
bgyori, line 59271: Is it at all possible to identify a specific person who could serve as contact?

nagutm: https://www.who.int/news-room/articles-detail/call-for-data-jecfa-95-meeting

This page lists the contact name for jecfa@who.int as Kim Petersen.

bgyori: I found kpetersen@who.int on a publication as a specific email address. Don't see an ORCID though.

Issue #1246, opened 2024-11-02
Register EPA hazardous waste numbers


bgyori:
Came across chemical mappings "in the wild" to what are called EPA Hazardous Waste numbers or RCRA (Resource Conservation and Recovery Act) waste numbers. To track these identifier mappings in a principled way, we'd need to have a prefix for this.

References:
- https://www.epa.gov/hw/defining-hazardous-waste-listed-characteristic-and-mixed-radiological-wastes#listed
- F codes: https://www.ecfr.gov/current/title-40/chapter-I/subchapter-I/part-261#261.31
- K codes: https://www.ecfr.gov/current/title-40/chapter-I/subchapter-I/part-261#261.32
- P/U codes: https://www.ecfr.gov/current/title-40/chapter-I/subchapter-I/part-261#261.33

bgyori:
Resolved in #1256

PR #1256, opened 2024-11-05
Add `epawaste` subprefixes

nagutm:
This pull request addresses the following issue: https://github.com/biopragmatics/bioregistry/issues/1246

The EPA hazardous waste numbers are divided into 4 distinct lists (F, K, P and U) that represent 3 different sematic spaces. As such I have curated 3 subprefixes under the parent prefix `epawaste`: `epawaste.flist`, `epawaste.klist`, `epawaste.pulist`.

Unlike https://github.com/biopragmatics/bioregistry/pull/1238, there is not a lot of shared information between these 3 sematic subprefixes and the parent prefix also does not represent any sematic space itself so I felt that curating them without a general `epawaste` parent prefix would be sufficient.


bgyori:
Actually, I think a single-prefix solution might make sense here. The different letters do have specific meanings but ultimately this is similar to how something like MeSH has multiple sub-trees for different types of entities or how GO consists of three main branches. When looking at mappings to this resource, we don't necessarily want to have to triage IDs based on the first letter and use a different prefix entirely.

src/bioregistry/data/bioregistry.json
```diff
@@ -35957,6 +35957,35 @@
...
+    "description": "This resource provides identifiers for listed hazardous wastes which are substances explicitly identified due to their potential harm to human health and the environment. These are categorized into four groups (F, K, P, and U) based on specific criteria, such as their source or inherent properties.",
+    "example": "F001",
+    "example_extras": [
+      "K001",
+      "P001",
+      "U001"
+    ],
+    "github_request_issue": 1246,
+    "homepage": "https://www.ecfr.gov/current/title-40/chapter-I/subchapter-I/part-261",
+    "name": "EPA Listed Wastes",
```
bgyori, line 35980:
```suggestion
    "name": "EPA Hazardous Wastes",
```

PR #1258, opened 2024-11-10
Fix branch reuse issue in new prefix PR creation workflow

nagutm:
The GitHub Actions workflow for registering new prefixes through the issue template is reusing the same branch for each new issue, leading to the content of the second pull request overwriting the first. This was first noticed in this PR: https://github.com/biopragmatics/bioregistry/pull/1240. This occurs because the workflow does not generate a unique branch for each new issue. As a result, subsequent PRs are force-pushing content to the previous PR's branch instead of creating new, independent branches for each PR.

This PR introduces the following changes to the `new_prefix_pr.yml` to address this issue:
- Create a step to extract the issue number from the issue URL.
- Dynamically generate a unique branch name (`create-pull-request/patch-<issue_number>`) for each new issue.
- Modify the PR creation step to use the dynamically generated branch name, ensuring that each PR is linked to its own unique branch.


bgyori:
This makes sense to me, and it seems to work via testing on a fork repository. One more thing that I think could be improved is that the update code is invoked even for issues that are not new prefix requests. For example, https://github.com/biopragmatics/bioregistry/actions/runs/11645684752/job/32428984849 technically, which is a non-new-prefix issue runs the same update code as a new prefix issue would. Is it possible to detect early whether something is not a new prefix issue and end the workflow under that condition before attempting to run an update?

bgyori:
Following up on my previous comment, given that we use
```yml
on:
  issues:
    types: [opened]
```
as the starting point for them action, it should be possible to add something like

```yml
jobs:
  get_issue_details:
    runs-on: ubuntu-latest
    steps:
      - name: Get issue details
        id: get_issue
        uses: actions/github-script@v6
        with:
          script: |
            <here we can access context.payload.issue to determine what is in the issue>
```
to examine the content/metadata for the issue and determine if it is relevant for a new prefix, before invoking the update code and the PR code.

nagutm:
I've updated the workflow to check if an issue contains the required `New` and `Prefix` labels before running the main script and creating a pull request. I've tested this in my forked repository and it successfully halts the workflow before the Update step.

.github/workflows/new_prefix_pr.yml
```diff
@@ -14,21 +14,47 @@ jobs:
...
+        with:
+          script: |
+            const issue = context.payload.issue;
+            const hasRequiredLabels = issue.labels.some(label => label.name === "New") && issue.labels.some(label => label.name === "Prefix");
+            core.setOutput("hasRequiredLabels", hasRequiredLabels ? 'true' : 'false');
+      - name: End Workflow if Labels are Missing
+        if: steps.check_labels.outputs.hasRequiredLabels == 'false'
+        run: |
+          echo "Issue does not have 'New' and 'Prefix' labels. Ending workflow."
+          exit 0
```
bgyori, line 31: If we exit the workflow here, is it still necessary to check below for 
```
if: steps.check_labels.outputs.hasRequiredLabels == 'true'
```
or can those be omitted?

nagutm: Omitting the conditionals does not actually halt the workflow from running for non-prefix issues if that is the goal.

Omitting conditionals:
image

Leaving conditionals in:
image

bgyori:
I think this looks good. The coverage check fails due to the few lines added to the Python code for the action. This is somewhat misleading since that code was not tested even before the PR, so flagging the diff as not tested is not exactly meaningful. Still I suppose adding a unit test for `process_new_prefix_issue` pointing to some old (closed) issue and checking the Resource returned could be useful anyway.

Issue #1257, opened 2024-11-08
Creative Biolabs - Recombinant Antibodies


garybrandam:
### Prefix

creativebiolabs-net

### Name

Creative Biolabs - Recombinant Antibodies

### Homepage

https://www.creativebiolabs.net

### Source Code Repository

bioregistry

### Description

[Creative Biolabs](https://www.creativebiolabs.net/) is a supplier of a wide range of antibodies from therapeutic antibodies and single domain antibodies to bispecific antibodies and intrabodies.

### Example Local Unique Identifier

TAB-884

### Contributor Name

garybrandam

### Contributor GitHub

garybrandam

### Contributor ORCiD

na

nagutm:
Hi @garybrandam, thank you for your contribution to the Bioregistry.  I have opened a pull request related to Creative Biolabs that you can monitor here: https://github.com/biopragmatics/bioregistry/pull/1260.

I was wondering if you could share some more information about yourself such as you full name, email address and ORCID number if applicable in order to properly credit you with this contribution in the Bioregistry. Thanks!

bgyori:
Added in #1260

PR #1260, opened 2024-11-11
Add prefix: creativebiolabs

nagutm:
This pull request addresses the following issue: https://github.com/biopragmatics/bioregistry/issues/1257

Creative Biolabs provides identifiers for various antibody related entities. Their website lists the following categories of products that have unique identifiers which I decided to break into 2 broad semantic spaces: 
- `creativebiolabs.antibody`: Recombinant antibodies, Functional antibodies, AbPlus, Formats, Research area, Engineered antibodies, AbProcessing
- `creativebiolabs.protein`: Tetramers, Proteolytic enzymes, Antibody-like Scaffold proteins, Therapeutic protein, Soluble T Cell receptor

I also considered further dividing these into 5 semantic subspaces to include `creativebiolabs.enzyme`, `creativebiolabs.processing` and `creativebiolabs.research` and even leaving it as a single all encompassing prefix. 

I was unable to track down a specific point of contact for this resource and have currently left in some placeholder information for the original contributor of this resource as well.


src/bioregistry/data/bioregistry.json
```diff
@@ -22211,6 +22211,58 @@
...
+  "creativebiolabs.antibody": {
+    "contact": {
+      "email": "info@creative-biolabs.com",
+      "name": "Creative Biolabs Team"
+    },
+    "contributor": {
+      "email": "placeholder@example.com",
+      "github": "garybrandam",
+      "name": "Gary",
+      "orcid": "0000-0000-0000-0000"
```
bgyori, line 22223: @garybrandam we require ORCIDs for contributions and can't proceed with merging the additions unless you can provide an ORCID. We also generally require emails for contributors. Please let us know and we can add it!

src/bioregistry/data/bioregistry.json
```diff
@@ -22211,6 +22211,58 @@
     "name": "Crates.io",
     "uri_format": "https://crates.io/crates/$1"
   },
+  "creativebiolabs.antibody": {
+    "contact": {
+      "email": "info@creative-biolabs.com",
+      "name": "Creative Biolabs Team"
```
bgyori, line 22217: @garybrandam the Bioregistry expects a specific point of contact person for each prefix. Could you provide contact info for an appropriate person, ideally name, ORCID, email, GitHub handle if available?

bgyori:
@cthoyt since @garybrandam is not responding but the resource is still worth adding, we were planning to switch to @nagutm as contributor and finalize the PR.

cthoyt (changes requested):
Please remove the contact - info@creative-biolabs.com - before merging, again, because the Bioregistry policy is only to include explicit single point of contacts.

bgyori:
> I'm skeptical about adding these since the contributor isn't responsive, there isn't a way to resolve these IDs, and I guess there isn't a way to get a full list of identifiers.
> 
> I'm curious what makes you think it's worth adding. Have you seen these identifiers appear in some other data resource? Is it valuable just for the fact that it addresses a community suggestion? Or maybe something else?

I don't have a strong feeling about it, it's just that (1) someone explicitly suggested it (2) the work of curating the resource is already done and (3) it appears to be a legitimate, and notable resource, in particular as a producer of antibodies, so being able to refer to their catalogue numbers via CURIEs in a principled way seems like a nice thing to have for the community in general.

> 
> There are various places where I've written notes on my thought process for what makes something worthy but I don't think I've prepare a centralized guide before.
> 
> If you can take care of deleting the group email then this is good enough for now.

PR #1263, opened 2024-11-19
Curate hscgd publication

nagutm:
This pull request curates a new prefix `hscgd` based on the following PubMed paper: https://pubmed.ncbi.nlm.nih.gov/39470706/
This resource provides identifiers for single-cell whole-genome sequencing data.

Each entry consists of a 'Project accession' that is resolvable using the uri_format and used as the primary identifier for this resource. In addition to that, there are separate pages with unique identifiers for 'Sample accessions' that correspond to each 'Project accession' and 'Run accessions' that correspond to each 'Sample accession'. The Sample and Run accession pages are directly accessible from the page for the corresponding 'Project accession'. I considered curating the Project, Sample and Runs as separate sematic spaces, however I felt that the Project accession alone was sufficient enough to describe the main idea of the resource.

The database also lists specific 'IDs' for each project that are numbered sequentially, however it does not make a lot of sense to me to use that as the primary identifier.


bgyori:
Upon more detailed review, it appears that the project identifiers here are actually resolving ones from other namespaces, in particular, `insdc.sra` and `ena.embl`. Overall, we might want to classify this as not notable.

PR #1269, opened 2024-11-21
Adding NFDI4Chem collection

StroemPhi:
This PR adds the NFDI4Chem collection to the Bioregistry.


cthoyt:
Hi @StroemPhi, thanks for the contribution! As you can see from the unit tests, there are a few prefixes listed here that aren't valid bioregistry prefixes:

- ssnsystem
- sosa
- oboe
- ssn
- nmrcv
- vibso

I can support you to add these prefixes, or you can move them into the description with some information that there's a bit of extra work to do.

StroemPhi:
All but vibso should be added imho. Let me have a try with ssnsystem and then I/we can see how good I manage, bofore I'll add the other ones. And VIBSO is not yet registered properly as it is still WIP. So that would be a case of prefix-parking, right?

cthoyt:
VIBSO is fine to add - because there's a public artifact at https://github.com/NFDI4Chem/VibrationalSpectroscopyOntology it's not considered prefix parking

cthoyt:
Note that ontokin is registered under https://bioregistry.io/registry/worldavatar.kin

cthoyt (approved):

PR #1282, opened 2024-11-25
Add new provider for pdb: dionysus

nagutm:
This pull request curates a new provider for PDB structure: `dionysus`.

PubMed: https://pubmed.ncbi.nlm.nih.gov/39436020/
Database: www.dsimb.inserm.fr/DIONYSUS/


src/bioregistry/data/bioregistry.json
```diff
@@ -86891,6 +86891,21 @@
...
+        "name": "DIONYSUS",
+        "publications": [
+          {
+            "doi": "10.1093/nar/gkae890",
+            "pubmed": "39436020",
+            "title": "DIONYSUS: a database of protein-carbohydrate interfaces",
+            "year": 2024
+          }
+        ],
+        "uri_format": "https://www.dsimb.inserm.fr/DIONYSUS/structure/$1"
```
bgyori, line 86908: I just tried this and one problem here is that the example ID listed for pdb ([2gc4](https://bioregistry.io/reference/pdb:/2gc4) doesn't resolve in this resource, so the example link shown on the web would be broken.

nagutm: This resource only appears to import a filtered set of PDB IDs that are carbohydrate-containing structures and match specific keywords related to the classification tag used by PDB. This explains why 2gc4 cannot be resolved with this uri_format.

I'm not sure how much value there is to curating a provider that cannot resolve all IDs for its parent resource. I think we could consider marking this as `not_notable` instead.

bgyori: Yeah, I don't think in itself it's a problem that not every ID is resolved, but I think overall not sure if it's worth adding this given the anticipation of  the broken example link.

nagutm: One other possible solution to this could be to curate this as its own prefix and use the "provides" key to signal that is it is a PDB provider. That way we could use an example ID that would create a working example link.

cthoyt:
@bgyori yes, let's create a separate issue and follow up with that there. For now, this PR is ready to merge.

bgyori:
Ironically, the website is completely down now, the month after publication...

PR #1286, opened 2024-11-26
Update mobidb prefix

nagutm:
This pull request curates a new publication for the `mobidb` prefix.
It also updates the `homepage` and `uri_format` based on the updated URL provided in the PubMed paper.

PubMed: https://pubmed.ncbi.nlm.nih.gov/39470701/


bgyori:
Looks good, thanks!

PR #1291, opened 2024-12-03
Mock PubMed API and update filtering logic in paper ranking model.

nagutm:
This pull request addresses two issues related to the paper ranking model and associated testing.

**1. Mocking `pubmed_client` to improve test speed and stability**
- Mocks the `pubmed_client` in the unit test to avoid API calls, improving test speed and stability. 
- A new file, `mock_pubmed_data.json`, has been added, containing mock data to simulate PubMed API responses for fetching metadata.

**2. Update filtering logic for curated publications**
- Modified the logic to exclude PubMed IDs from curated_papers_df, publication_df, and curation_df when fetching new PubMed papers, ensuring no duplicates from already curated or published entries.


nagutm:
The `mock_to_csv` call count test was removed because it is not entirely necessary for the purpose of this smoke test, which primarily ensures that the pipeline runs successfully without errors.

I think that mocking `to_csv` globally captured all indirect uses of `to_csv` outside the main function. This resulted in higher than expected call counts.

src/bioregistry/analysis/paper_ranking.py
```diff
@@ -397,7 +397,9 @@ def main(bioregistry_file: Path, start_date: str, end_date: str) -> None:
     importances_df.to_csv(importance_path, sep="\t", index=False)
 
     # These have already been curated and will therefore be filtered out
-    curated_pmids = set(curated_papers_df["pubmed"])
+    curated_pmids = set(curated_papers_df["pubmed"]).union(
+    publication_df["pubmed"], curation_df["pubmed"]
```
cthoyt, line 401: can you look into why CI isn't picking up on this formatting error?

src/bioregistry/data/mock_pubmed_data.json
```diff
@@ -0,0 +1,22 @@
+{
```
cthoyt, line 1: stick this in the tests/ directory - this isn't necessary for people who install the code

src/bioregistry/analysis/paper_ranking.py
nagutm: I'm not entirely sure. I can't find anything in the logs that would explain this. I did recently switch to using a new laptop recently which has been causing a wide range of issues but I don't think anything in my local environment should affect the CI from picking up formatting errors?

bgyori: I found this surprising as well, don't see any obvious explanation. In any case, the indentation fixed as far as this PR goes.

PR #1294, opened 2024-12-06
Add provider for metacyc prefix and consider new prefix for unique IDs

nagutm:
PubMed: https://pubmed.ncbi.nlm.nih.gov/39558185/
Database: [PlantCyc](https://pmn.plantcyc.org/organism-summary?object=PLANT&_gl=1*v0eipa*_ga*MjYwMjg2NzE0LjE3MzMyNzg2Mjk.*_ga_924HD6E7CQ*MTczMzQ1NzEzOS40LjEuMTczMzQ1NzYxMS4wLjAuMA..)

This resource has 4 different sematic spaces: Compunds, Reactions, Pathways and Gene/Proteins

The first 3 use the same identifier information from `metacyc.compound`, `metacyc.pathway` and `metacyc.reaction`. They are curated as providers for these resources respectively.

The Gene/Protein subspace seems to have a mix of IDs used. 
- Some of them are [TAIR](https://www.arabidopsis.org/) IDs , however the current resolver links on the `tair` pages in the Bioregistry seem to be outdated and not resolving IDs as expected so I have not curated anything under this prefix until I fix those. [example of a tair ID](https://pmn.plantcyc.org/gene?orgid=PLANT&id=AT5G47720)
- The other category of IDs seem to be unique to this resource with a general pattern of ^G-\\d+$. These could be curated as a separate prefix. [example of a unique ID](https://pmn.plantcyc.org/gene?orgid=PLANT&id=G-14542)


bgyori:
We can get back to this once we know more about #1335

nagutm:
Because TAIR’s site redesign broke legacy accession URLs for genes and loci and because the G‑\d+ IDs would require a separate new prefix, this introduces too many moving parts to merge cleanly in this PR. Additionally, the compound, pathway and reaction namespaces do not always use the corresponding metacyc identifier format as I originally thought which makes the curation of these providers dubious. I’m going to close this PR for now, and we can potentially revisit them when there is some more clarity on these issues.

PR #1295, opened 2024-12-07
Add prefix: npatlas

nagutm:
PubMed: https://pubmed.ncbi.nlm.nih.gov/39588755/

This pull request curates a new prefix `npatlas` based on the PubMed paper above. 
This resources provides identifiers for microbially derived natural products.


bgyori:
This looks good @nagutm! Same as for StreptomeDB, this resource has a couple of prior publications that we might as well curate on this PR.

PR #1312, opened 2024-12-12
New prefix: biotop

cthoyt:


src/bioregistry/data/bioregistry.json
```diff
@@ -10286,6 +10286,45 @@
       }
     }
   },
+  "biotop": {
+    "contact": {
+      "email": "stefan.schulz@medunigraz.at",
+      "github": "steschu63",
+      "name": "Stefan Schulz ",
```
cthoyt, line 10293: space in name, should add a more generic test against leading/trailing whitespaces

PR #1330, opened 2024-12-14
Add prefix: markerdb

nagutm:
PubMed: https://pubmed.ncbi.nlm.nih.gov/39535054/
Database: https://markerdb.ca/

This pull request curates a new prefix `markerdb` which provides identifiers for various molecular biomarkers.

This resource is curated without a `uri_format` since the number in the URL for individual entries seems irrelevant and unrelated to the MarkerDB ID associated with each biomarker.


bgyori (approved):
I fixed the curations table which had the wrong PMID in it, added the Github handle for the contact, and extended the description.

PR #1346, opened 2025-01-06
Add additional ICD11 providers

cthoyt:
This PR clarifies that there are several URLs that can resolve for ICD11 codes. It also fixes the ICF URI format string, since it was overlapping and too generic with ICD11.

This PR **does not** depend on a discussion about the difference between ICD11 numeric identifiers and other codes, like `9D00.2`, that appear in the MMS browser.

Using _Astigmatism_ (`icd11:402558626`) as an example, here are 4 different ways to browse the same numeric identifier:

| Description  | URI Format | Example |
|--------------|------------|---------|
| API Endpoint | `http://id.who.int/icd/entity/$1` | http://id.who.int/icd/entity/402558626 |
| Default Browser (redirects to Mortality and Morbidity Statistics) | `https://icd.who.int/browse11/l-m/en#/http://id.who.int/icd/entity/$1` | https://icd.who.int/browse11/l-m/en#/http://id.who.int/icd/entity/402558626 |
| Foundation Browser  | `https://icd.who.int/browse/2024-01/foundation/en#$1` | https://icd.who.int/browse/2024-01/foundation/en#402558626 |
| Mortality and Morbidity Statistics Browser  | `https://icd.who.int/browse/2024-01/mms/en#$1` | https://icd.who.int/browse/2024-01/mms/en#402558626 |

This mostly supersedes #1345, but does not add a prefix synonym for "foundation". This can be considered in a different PR.

cc @sierra-moxon


src/bioregistry/data/bioregistry.json
```diff
@@ -54503,7 +54526,7 @@
     },
     "name": "International Classification of Functioning, Disability and Health",
     "pattern": "^\\d+$",
-    "uri_format": "http://id.who.int/icd/entity/$1"
+    "uri_format": "https://icd.who.int/dev11/l-icf/en#/http://id.who.int/icd/entity/$1"
```
matentzn, line 54529: Looks weird?

src/bioregistry/data/bioregistry.json
```diff
@@ -53853,6 +53853,29 @@
...
+    "providers": [
+      {
+        "code": "api",
+        "description": "This goes to an API endpoint. For more information, see https://icd.who.int/icdapi",
+        "homepage": "https://icd.who.int",
+        "name": "ICD 11 API Endpoint",
+        "uri_format": "http://id.who.int/icd/entity/$1"
+      },
+      {
+        "code": "foundation",
```
matentzn, line 53865: Does this mean we will get icd11.foundation as a prefix?

PR #1348, opened 2025-01-10
New prefix: `theiler`

cthoyt:


bgyori:
Not sure what the context was for the PR but it's worth noting that Theiler Stages are also represented in emouseatlas.org at https://www.emouseatlas.org/emap/ema/theiler_stages/StageDefinition/stagedefinition.html, which I recall from when I was looking into that resource.

PR #1350, opened 2025-01-10
Commit Predictions File with Git Hash Link, Remove Artifact Upload

nagutm:
This PR refactors the GitHub Actions workflow for the paper ranking script to ensure the predictions file is committed directly to the repository. The following changes were made:

1. Removed Artifact Upload Step:
- The actions/upload-artifact step in the workflow is currently used to store files generated during the workflow. These artifacts are then available for download in the GitHub Actions interface under the "Artifacts" section for that workflow run.
- The addition of the commit and push step below makes this step unnecessary and was removed to avoid redundant storage.

2. Added Commit and Push Step:
- Introduced a new step to commit and push the predictions file directly to the `exports/analyses/paper_ranking/` directory in the repository.
- This ensures that the predictions file is saved in the repository for future use if necessary.

Testing via a forked repository successfully showed that the predictions file was saved to the correct directory once the workflow run was complete.


.github/workflows/paper_ranking.yml
```diff
@@ -41,11 +45,16 @@ jobs:
...
 
-    - name: Upload Full List as Artifact
-      uses: actions/upload-artifact@v3
-      with:
-        name: full-predictions-list-${{ env.START_DATE }}-to-${{ env.END_DATE }}
-        path: exports/analyses/paper_ranking/predictions_${{ env.START_DATE }}_to_${{ env.END_DATE }}.tsv
+    - name: Configure Git
+      run: |
+        git config user.name "github-actions[bot]"
+        git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
cthoyt, line 51: what does `41898282` mean?

bgyori:
> I would suggest appending to the existing predictions file, rather than creating new ones each time. If you need to keep track of date provenance for this, then you could add a new column for the publication date

Here is why we designed this process as it is currently:
- The ranking is incremental in that it combines new papers with uncurated past papers to re-rank all papers. So it's not a simple append operation. This ensures that uncurated past papers that are highly ranked can still be surfaced.
- In addition, we wanted to ensure that each month's comment posting new ranking results points to a specific file that is associated with that comment, even if newer comments/rankings have been posted since.

I do have an idea that could work and serve as a better solution:
- We could have a single version controlled ranking file
- Each time the ranking is done, the ranking file is overwritten with the new rankings in a github-actions commit
- The comment posted on the issue doesn't simply add a link to this file from the `main` branch but uses the github-actions commit's hash (the one added in the previous step) in the direct link to the file. This ensures that the link will continue to exist later while also being reproducibly associated with the current issue comment.

Should we try this?

.github/workflows/paper_ranking.yml
nagutm: The number is a unique identifier for the GitHub Actions bot account. It ensures that commits made by GitHub Actions workflows are clearly attributed to the bot.

.github/workflows/paper_ranking.yml
```diff
@@ -41,11 +45,20 @@ jobs:
...
+
+    - name: Commit and Push Changes
+      run: |
+        git add exports/analyses/paper_ranking/predictions.tsv
+        git commit -m "Add predictions file for ${{ env.START_DATE }} to ${{ env.END_DATE }}"
```
cthoyt, line 56: I identified a pretty significant bug in the way that the start date and end date affect the way the paper ranking workflow work and left some comments on it in https://github.com/biopragmatics/bioregistry/pull/1358. Please take a look at that PR before finishing this one, since fixing this bug has implications for how the configuration in this PR should be changed

cthoyt (changes requested):

.github/workflows/paper_ranking.yml
```diff
@@ -43,11 +47,20 @@ jobs:
...
```
bgyori, line 58:
```suggestion
        git commit -m "Update predictions file with papers between ${{ env.START_DATE }} and ${{ env.END_DATE }}"
```

.github/workflows/paper_ranking.yml
bgyori: @nagutm will fix the date range issue in the ranking script on a separate PR as a follow up to this.

bgyori:
@nagutm before we wrap this up: is adding the file `exports/analyses/paper_ranking/predictions_2024-12-11_to_2025-01-10.tsv` necessary here? I imagine given the change to a single `predictions.tsv` we would want to end up with just that one file (and get rid of the date-specific ones) after this PR is merged but I might be missing something.

nagutm:
> @nagutm before we wrap this up: is adding the file `exports/analyses/paper_ranking/predictions_2024-12-11_to_2025-01-10.tsv` necessary here? I imagine given the change to a single `predictions.tsv` we would want to end up with just that one file (and get rid of the date-specific ones) after this PR is merged but I might be missing something.

The date specific prediction file was added before the suggestion to use just one predictions file. It can be safely removed.

bgyori:
Alright, this looks good. The code coverage warning is an artifact of how coverage rules are applied whereby for #1358 coverage wasn't run whereas here it was - choosing to ignore this.

PR #1351, opened 2025-01-10
Curating ranked papers for relevance

nalikapalayoor:
This pull request adds the curations for papers that I have found to be irrelevant to the Bioregistry. These papers came from the potentially relevant paper ranking table. I will curate relevant entries in separate PRs.


bgyori:
Thanks @nalikapalayoor! One minor question also to @nagutm is that I think most of the irrelevant entries are fundamentally because a paper is not about an identifiers resource (or about a provider for an identifiers resource). Are we overusing the `irrelevant_other` tag given that we also have two more specific tags, `not_identifiers_resource` and `no_website`?

image
 at https://biopragmatics.github.io/bioregistry/curation/literature

bgyori:
Also, after we clarify the curation tags, we should also lint this file to make sure it's sorted, as the test suggests

nagutm:
I think that the `not_identifier_resource` definition might be too restrictive and could be expanded to include papers describing software repositories, data visualization tools, and any other externally linked 'resource' that falls along these lines. Currently these types of papers are being classified as `irrelevant_other` per the current definitions. 

I think that the `no_website` tag can be modified to `non_resource_paper`. This would describe any papers that are self-contained and don't describe/link to any externally developed resources.

The `irrelevant_other` tag can be retained as a catch-all that would cover anything else. 

I think these revised definitions/tags would help to make the distribution of tags for irrelevant papers more balanced and less one-sided as they currently are.

bgyori:
@nagutm is it straightforward to make these changes to the table overall (i.e., past `irrelevant_other` tags that might fall under the definition of some of the more specific categories)? If so, we could do that.

nagutm:
Changing the tags for all previous papers tagged as `irrelevant_other` would require us to manually check each of the papers and re-classify them. It's doable but might take a little time. I'll update the guide documentation and CurationRelevance vocabulary files before updating the `curated_papers.tsv` file.

nalikapalayoor:
Hi, thank you for looking at this! I can go in and reclassify the papers associated with this PR based on those redefined tags. I will also relint the file. I will make these changes and push the changes later today!

nalikapalayoor:
I have updated the tags for the papers determined to be irrelevant based on the new definitions discussed:

not_identifiers_resource: papers with software repos, data visualization tools, externally linked resources, as well as databases not related to defining new identifiers

non_resource_papers: self-contained papers that don't link to external resources

irrelevant_other: all other irrelevant papers

I also linted the currated_papers.tsv file after updating these tags.

bgyori:
This looks good, we just need to update all the places where the `no_website` tag is referred to, to use the new `non_resource_papers` name, and update any changes to the definition.

nagutm:
See https://github.com/biopragmatics/bioregistry/pull/1359 for these updated changes

bgyori (approved):

PR #1352, opened 2025-01-11
Various updates to ICD 11 prefix record

matentzn:
I want to say that this is a compromise I can just tolerate, but I don't really like it. The choice of having `icd11`, `ICD-11` and `ICD11` be associated with the foundation component strikes me as questionable (but I also don't know _for sure_ if this is wrong, else I would insist). What I really do not like is to have `icd11` being the default prefix for the icd11 foundation, rather than the much less ambiguous `icd11.foundation`.

I would say, lets not do any more damage now, this PR will satisfy our immediate needs, but IMO we should just have the icd11 team itself decide what they want. I know them all quite well, and will reach out now to see what they say.

### DIFF
--- a/exports/alignment/biolink.tsv
+++ b/exports/alignment/biolink.tsv
@@ -7,7 +7,6 @@ DOID-PROPERTY	http://purl.obolibrary.org/obo/doid#$1	False	True
 gff3	https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md#$1	False	False
 GOP	http://purl.obolibrary.org/obo/go#$1	False	True
 gtpo	https://rdf.guidetopharmacology.org/ns/gtpo#$1	False	False
-icd11.foundation	http://id.who.int/icd/entity/$1	False	False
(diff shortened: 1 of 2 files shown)


matentzn:
Need help with that error!

src/bioregistry/data/bioregistry.json
```diff
@@ -53913,7 +53916,7 @@
       "name": "Charles Tapley Hoyt",
       "orcid": "0000-0003-4423-4370"
     },
-    "description": "ICD11 codes, which typically appear in health records and can be used for post-coordination",
+    "description": "An ICD-11 code is a standardized alphanumeric identifier used to classify and document diseases, conditions, and health-related issues for global healthcare and statistical purposes.",
```
cthoyt, line 53919: can you please give some more context?

Who uses ICD11 codes, and where? Why do they exist as a counterpart to ICD11 identifiers, which have URIs and everything, but these don't?

cthoyt:
@matentzn I added a comment for follow-up, I think that it would be a missed opportunity not to make the descriptions much more detailed and useful for non-ICD-insiders.

PR #1353, opened 2025-01-11
Remove OBCI OBO PURL

### LINKED ISSUE #1347: Update Ontology for Biomarkers of Clinical Interest
### Prefix

obci

### Explanation

It is no longer clear if this resource will be under the umbrella of the OBO Foundry, and thus the PURLs will need to be changed to something else. There is not yet a resolvable home for the individual terms. What is the best way forward to deal with this? The easiest I can think of is to remove the URI example from the Providers table, but I'm not sure if this is allowed. I'd hate to have the whole prefix pulled, but if that's the only solution then so be it.

### Contributor ORCID

 0000-0001-5809-9523

bgyori:
This PR removes the OBO PURL from the OBCI record, as discussed in #1347 without removing anything else from the otherwise valid record. Fixes #1347.


cthoyt (changes requested):
Please hold off on making this change until https://proteininformationresource.org/staff/nataled/OBCI/obci.owl is updated to not use this URI format. Otherwise, it's not possible to parse the URIs in this file.

bgyori:
@nataled please look into this and let us know if you will change the OWL file on your end or if we should pursue a different solution.

bgyori:
@nataled do you have any update/comment on this issue?

nataled:
We will be changing the URIs, but we're still discussing what URIs will be used going forward. Hopefully we'll decide at our next meeting in a few weeks.

bgyori:
OBCI just came up at the Biocuration conference so I looked back at this entry :) Looks like the OWL file still has the OBO PURLs in it so there is no resolution to the issue here yet.

cthoyt:
This hasn't been updated in 18 months, so I will close the PR to reduce clutter in this repo.

@nataled please feel free to comment on the original issue if/when the status of OBCI changes

PR #1358, opened 2025-01-13
Refactor paper ranking workflow

cthoyt:
This PR refactors the paper ranking workflow with the following:

1. Remove implementation from inside CLI definitions. This makes it easier to test (rather than having to test that a click command works, you run the function directly)
2. Add type annotations
3. Reduce duplicated code
4. Reuse itertools when appropriate
5. Better function names
6. Hide INDRA imports inside relevant functions so it can be directly mocked
7. Add several TODOs and FIXMEs, but these are for future work. Right now, I just want to check that this first refactor still works the same way. 

It also does:

1. Define optional requirements for the paper tranking workflow in pyproject.toml instead of in an ad-hoc requirements.txt file
2. Redefine the paper ranking workflow to install with the optional from pyproject.toml
3. Add a tox environment for running the paper ranking workflow, so it's reproducible by anyone
4. Reuse builtin paths in the Bioregistry module during tests


bgyori:
I think restructuring the code is okay, without going into discussions about detailed choices, perhaps @nagutm could check to see if there are any unintended changes to the overall behavior we need to address. I think we should look into the reldate / explicit start and end date discrepancy on a separate PR to not mix refactoring with that (That issue happens to not make a practical difference in the monthly runs but is generally something we should fix.)

nagutm (commented):
I tested the basic functionality with the refactored code and there does not appear to be any unexpected changes to the general workflow. I think we can merge this PR in order to address the issue of the start and end dates.

PR #1365, opened 2025-01-17
Update prefix: seed --> seed.subsystem

### LINKED ISSUE #1339: Update SEED Subsystem
### Prefix

seed

### Explanation

The `seed.subsystem` synonym should be removed and used to create it's own entry as a prefix. This follows the convention of having sub-prefixes that are indicative of the data type. For example there are a `seed.compound` and `seed.reaction` entries, and I will be shortly proposing a new `seed.role` entry.

### Contributor ORCID

0000-0002-7674-5194

nalikapalayoor:
After reviewing the current prefixes associated with The Seed, we decided to update the name of the 'seed' prefix to be 'seed.subsystems', and have 'seed' be a synonym for it instead. This PR makes this update, and also updates the 'uri_format' for 'seed.subsystems' to be a working url, as the old one seems to be outdated. 

Fixes #1339


src/bioregistry/data/bioregistry.json
```diff
@@ -104533,6 +104460,79 @@
       "uri_format": "http://modelseed.org/biochem/reactions/$1"
     }
   },
+  "seed.subsystems": {
```
bgyori, line 104463:
```suggestion
  "seed.subsystem": {
```

src/bioregistry/data/bioregistry.json
```diff
@@ -104533,6 +104460,79 @@
...
+        "github": "samseaver",
+        "name": "Sam Seaver",
+        "orcid": "0000-0002-7674-5194"
+      }
+    ],
+    "go": {
+      "description": "Project to annotate the first 1000 sequenced genomes, develop detailed metabolic reconstructions, and construct the corresponding stoichiometric matrices",
+      "homepage": "http://www.theseed.org",
+      "name": "The SEED;",
+      "prefix": "SEED.SUBSYSTEMS",
```
bgyori, line 104485: @nalikapalayoor we shouldn't make any changes to anything inside the keys for the mapped prefixes, so e.g., the line above should stay the same as above.

src/bioregistry/data/bioregistry.json
bgyori: We should use a singular form here

src/bioregistry/data/bioregistry.json
```diff
@@ -104533,6 +104460,79 @@
...
+      "uri_format": "http://www.theseed.org/linkin.cgi?id=$1"
+    },
+    "mappings": {
+      "biocontext": "SEED.SUBSYSTEMS",
+      "go": "SEED.SUBSYSTEMS",
+      "miriam": "seed.subsystems",
+      "n2t": "seed.subsystems",
+      "ncbi": "SEED.SUBSYSTEMS"
```
bgyori, line 104493: Same here: all the mapped values in these external registries should remain unchanged.

bgyori:
The diff looks weird because of the change in alphabetical order between prefixes but to summarize, the changes are:
- Rename `seed` to `seed.subsystem`
- Change the existing synonym from `seed.subsystem` to `seed`
- Add an override `uri_format`

PR #1366, opened 2025-01-18
Add prefix: eccDNA

nalikapalayoor:
This pull request curates a new prefix eccDNA, which is a database based on single-cell whole genome sequencing data. This PR adds necessary metadata to bioregistry.json and adds the curated paper to currated_papers.tsv

https://pubmed.ncbi.nlm.nih.gov/39693568/


src/bioregistry/data/bioregistry.json
```diff
@@ -30588,6 +30588,35 @@
     "pattern": "^\\d{7}$",
     "uri_format": "http://purl.obolibrary.org/obo/ECAO_$1"
   },
+  "eccDNA": {
```
bgyori, line 30591: For new prefixes, the convention is to use all lowercase, but if capitalization is important to preserve in some form, then it can be done via synonyms (see also https://github.com/biopragmatics/bioregistry/blob/main/docs/CONTRIBUTING.md#minimum-new-prefix-requirements)

src/bioregistry/data/bioregistry.json
```diff
@@ -30588,6 +30588,35 @@
...
+      "name": "Zhiyun Guo",
+      "orcid": "0000-0003-2680-785X"
+    },
+    "contributor": {
+      "email": "palayoor.n@northeastern.edu",
+      "github": "nalikapalayoor",
+      "name": "Nalika Palayoor",
+      "orcid": "0009-0008-8406-631X"
+    },
+    "description": "scEccDNAdb is a database of more than 3.1 million single cell single cell eccDNA entries of human and mouse samples, each with a unique identifiers and features such as oncogenes, enhancers, and structural variants. It also features tools for analyzing and visualizing eccDNA data.",
```
bgyori, line 30603:
```suggestion
    "description": "scEccDNAdb is a database of single cell eccDNA entries of human and mouse samples, each with a unique identifier and features such as oncogenes, enhancers, and structural variants..",
```

src/bioregistry/data/bioregistry.json
```diff
@@ -30627,6 +30627,41 @@
...
+      "github": "nalikapalayoor",
+      "name": "Nalika Palayoor",
+      "orcid": "0009-0008-8406-631X"
+    },
+    "description": "scEccDNAdb is a database of single cell eccDNA entries of human and mouse samples, each with a unique identifier and features such as oncogenes, enhancers, and structural variants.",
+    "example": "ecDNA_hsa_01",
+    "github_request_issue": 1366,
+    "homepage": "https://lcbb.swjtu.edu.cn/scEccDNAdb/",
+    "name": "Single Cell ECC DNA Database",
+    "pattern": "^ecDNA_[a-z]+_\\d+$",
```
cthoyt, line 30647: it's surprising that the resource is called `eccdna` but the identifier has `ecDNA` with one fewer `c`

bgyori: Actually, I realize now that it's even more complicated. The example ID is valid 
image

but looking further, there are actually other ID's that use ecc as a prefix:

image

😮 

So the regex pattern will need to be slightly adjusted and we can add an extra example.

src/bioregistry/data/bioregistry.json
```diff
@@ -30627,6 +30627,41 @@
...
+      "email": "palayoor.n@northeastern.edu",
+      "github": "nalikapalayoor",
+      "name": "Nalika Palayoor",
+      "orcid": "0009-0008-8406-631X"
+    },
+    "description": "scEccDNAdb is a database of single cell eccDNA entries of human and mouse samples, each with a unique identifier and features such as oncogenes, enhancers, and structural variants.",
+    "example": "ecDNA_hsa_01",
+    "github_request_issue": 1366,
+    "homepage": "https://lcbb.swjtu.edu.cn/scEccDNAdb/",
+    "name": "Single Cell ECC DNA Database",
```
cthoyt, line 30646: please write out non-obvious acronyms like ECC in full in the title

src/bioregistry/data/bioregistry.json
```diff
@@ -30627,6 +30627,41 @@
...
+      "name": "Zhiyun Guo",
+      "orcid": "0000-0003-2680-785X"
+    },
+    "contributor": {
+      "email": "palayoor.n@northeastern.edu",
+      "github": "nalikapalayoor",
+      "name": "Nalika Palayoor",
+      "orcid": "0009-0008-8406-631X"
+    },
+    "description": "scEccDNAdb is a database of single cell eccDNA entries of human and mouse samples, each with a unique identifier and features such as oncogenes, enhancers, and structural variants.",
```
cthoyt, line 30642: what is an eccDNA? why do we care about them?

src/bioregistry/data/bioregistry.json
bgyori: For future reference, there are two types of entries: ecDNA (extrachromosomal DNA) and eccDNA (extrachromosomal circular DNA) so the ID pattern makes sense. I added an extra example to show both types.

bgyori: There is a problem: for the `eccDNA_*` case:
```
            for example in entry.example_extras:
                with self.subTest(prefix=prefix, identifier=example):
>                   self.assertEqual(entry.standardize_identifier(example), example)
E                   AssertionError: 'hsa_35832' != 'eccDNA_hsa_35832'
E                   - hsa_35832
E                   + eccDNA_hsa_35832
```
This has to do with some core code in `standardize_identifier` that automatically strips off the prefix if followed by a `_` at https://github.com/biopragmatics/bioregistry/blob/36cad81452322bf9fffb16e307662db3ce7fbc56/src/bioregistry/schema/struct.py#L1965-L1966. This might have to be changed since in this case we wouldn't want to apply it.

cthoyt: Set `"banana_peel": "",` to get around this

bgyori: Unfortunately `_` is hard-coded as an additional potential banana peel here:
https://github.com/biopragmatics/bioregistry/blob/36cad81452322bf9fffb16e307662db3ce7fbc56/src/bioregistry/schema/struct.py#L1960 so that won't work by itself.

PR #1369, opened 2025-01-19
Add test for duplicate PubMed in literature curation sheet

cthoyt:
I noticed `bioregistry lint` was causing some rows to get deleted, then noticed it was because there were duplicate curations. This PR adds a test for duplicate curations and in the five instances where there were duplicates, I deleted the later curated row. None of these affected the outcome, but there were conflicts in the calls from the curators.

@bgyori were you interested in doing inter-curator agreement statistics later? if not, then we can safely delete these duplicate curations


bgyori:
It's okay to remove the duplicates, they weren't planned to be curated again, and aren't needed in general. I just made small change to keep the more specific tag in the retained instance of each duplicate curation.

PR #1373, opened 2025-01-21
Add prefix: dome

nalikapalayoor:
This pull request curates a new prefix 'dome', which is a set of papers related to machine learning related publications. Machine learning has lots of applications in life sciences, which is why I decided it would be relevant to the bioregistry. This PR updates the json and tsv files. 

https://pubmed.ncbi.nlm.nih.gov/39661723/

PR #1374, opened 2025-01-21
Add prefix: dome

nalikapalayoor:
This pull request curates a new prefix 'dome', which is a set of papers related to machine learning related publications. Machine learning has lots of applications in life sciences, which is why I decided it would be relevant to the bioregistry. This PR updates the json and tsv files. 

This PR is the updated version of #1373 but with correct branching. 

https://pubmed.ncbi.nlm.nih.gov/39661723/


src/bioregistry/data/bioregistry.json
```diff
@@ -28681,6 +28681,35 @@
     },
     "name": "Descriptive Ontology for Linguistic and Cognitive Engineering"
   },
+  "dome": {
+    "contact": {
+      "email": "silvio.tosatto@unipd.it",
+      "name": "Silvio Tosatto",
+      "orcid": "0000-0003-4525-7793"
```
bgyori, line 28688: We can add `silvio-tosatto` as a Github handle as well

bgyori:
In virtually all situations, branch related issues can be fixed (using some combination of e.g., rebase, merge, reset, force push etc.) without having to start a new PR and it is preferable to stay on one PR once it's open. Still, for the time being, let's continue with this one.

Issue #1267, opened 2024-11-20
Should new prefix issue workflow logic consider one or more issues?


bgyori:
Currently (as well as before the changes in https://github.com/biopragmatics/bioregistry/pull/1258), the code turning new prefix issues into PRs iterates over multiple issues and thereby considers multiple prefixes simultaneously for addition. An example of this is #1266 where both #1257 and #1265 were processed and added to a PR.

Should this logic be preserved, or alternatively, should we look at the payload in this section
https://github.com/biopragmatics/bioregistry/blob/main/.github/workflows/new_prefix_pr.yml#L24-L26
to figure out what specific new issue triggered the workflow, and just create a PR from that given issue, not other past issues?

nagutm:
Based on the last line of this comment: https://github.com/biopragmatics/bioregistry/pull/1193#pullrequestreview-2354586596, I have been creating individual PRs for every new curation in order to limit the scope and not create too much cross communication between unrelated resources where information can easily get lost.

If we want to continue this then it probably makes more sense to limit the workflow to not iterate over every open issue. Especially considering that we often make edits and suggestions to the content of the new entry before merging it.

bgyori:
The single-issue-based workflow was implemented in #1376 and seems to have worked successfully for #1381.

PR #1376, opened 2025-01-22
Update New Prefix Workflow to Process Specific Issue

nagutm:
This pull request addresses the following issue: https://github.com/biopragmatics/bioregistry/issues/1267

The purpose of this PR is to update the new prefix issue workflow logic to only process the issue that triggered the workflow and not iterate over every relevant open issue as it currently does. I've outlined the changes below:

`github_client.py`:
- Added `get_form_data_for_issue` function to retrieve and parse form data for a specific issue

`new_prefix.py`:
- Updated the main function to accept an optional issue parameter
- Uses `get_form_data_for_issue` to only process information about the prefix corresponding to the specific issue number
- Only searches all relevant new prefix issues if issue parameter not provided

`new_prefix_pr.yml`:
- Added logic to extract issue number and run workflow only on that specific issue

Thanks to @bgyori for writing most of the logic to handle these changes in the `new_prefix_single` branch
I have tested the logic in a forked repository which shows that the updated workflow successfully generated a PR when a specific issue number is provided as a parameter


src/bioregistry/gh/new_prefix.py
```diff
@@ -228,40 +231,55 @@ def main(dry: bool, github: bool, force: bool):
...
-    issue_to_resource = {
-        issue_id: value
-        for issue_id, value in issue_to_resource.items()
-        if issue_id not in pulled_issues
-    }
-
-    if issue_to_resource:
-        click.echo(f"Adding {len(issue_to_resource)} issues after filter")
+    # If an issue number is given, process that specific issue
+    if issue is not None:
```
cthoyt, line 235: since both branches of this if statement return an `issue_to_resource` dictionary, can you please refactor both branches into their own functions

src/bioregistry/gh/new_prefix.py
```diff
@@ -215,10 +215,13 @@ def make_title(prefixes: Sequence[str]) -> str:
 @click.command()
 @click.option("--dry", is_flag=True, help="Dry run - do not create any PRs")
 @click.option("--github", is_flag=True, help="Use this flag in a GHA setting to set run variables")
+@click.option("--issue", is_flag=False, help="Specific issue to process rather than finding all relevant ones")
```
cthoyt, line 218: `is_flag=False` is the default - you can make this more explicit by setting `type=int`

```suggestion
@click.option("--issue", type=int, help="Specific issue to process rather than finding all relevant ones")
```

nagutm:
I have updated the `test_new_prefix.py` file to include two additional tests.
- `test_specific_issue` checks that the new workflow runs without error when a specific issue parameter is provided.
- `test_all_relevant_issues` checks that the old workflow runs without error when no issue parameter is provided and processes all relevant issues.

I also refactored some of the test prefix data to a file called `mock_prefix.py` in the `tests/resources` directory. There seems to be an issue with the GitHub testing environment being unable to recognize the import statement for this file. 

I'm not sure if placing the test prefix data in the `tests/resources` directory is the best option. Could I get some feedback on this and also how to potentially resolve the import issue?

bgyori:
Thanks @nagutm! I think we should turn the content of tests/resources/mock_prefixes.py into two separate JSON files in the tests/resources folder and then read these in test_new_prefix.py. This is better both in terms of making imports a non-issue and that static data structures are better maintained as data files rather than code.

bgyori (approved):

Issue #1335, opened 2024-12-17
TAIR resolver doesn't work


bgyori:
While reviewing https://github.com/biopragmatics/bioregistry/pull/1294 I noticed that the resolver for https://bioregistry.io/registry/tair.gene doesn't work anymore using either the direct URL to the provider (http://arabidopsis.org/servlets/TairObject?accession=Gene:2200934) or any of the metaregistry-based providers. 

![image](https://github.com/user-attachments/assets/7e68c648-4308-447d-9e67-e040ae7cc91c)

The current website for TAIR lists an accession number for each gene but the URL by which you get to that page has a "key" parameter which is a different number. Still, trying to fix the resolver URL by using 2200934 as the key doesn't work. Does anyone know what the relationship is between the TAIR accession numbers and the keys that appear in URLs?

nagutm:
On May 6th, 2024 TAIR announced that they are launching their new website. This new website affected the `tair.gene`, `tair.locus` and `tair.protein` prefixes. 

For `tair.gene` and `tair.locus`, the new website preserves the 'TAIR accession' in the details page for each entry, however it does not seem possible to use this TAIR accession in any kind of URL to go to the page of the corresponding entry. The 'key' parameter in the current URL also appears to be unrelated to the old TAIR accession.

 For `tair.protein`, the 'key' parameter in the current URL matches the TAIR accession exactly. The example below shows that the new website can resolve the TAIR accession number when plugged into the 'key' parameter.

 - old uri_format: http://arabidopsis.org/servlets/TairObject?accession=AASequence:1009107926
 - new uri_format: https://www.arabidopsis.org/protein?key=1009107926
 
 The `tair.protein` prefix can easily be updated with the current uri_format from the new website, but I'm not sure if the other two prefixes should have their identifiers updated from the old 'TAIR accession' to the new 'key' parameter.

PR #1384, opened 2025-01-28
New provider: athribonc

nalikapalayoor:
This PR adds a new TAIR identifier: 'athribonc'
https://pubmed.ncbi.nlm.nih.gov/39689042/


src/bioregistry/data/bioregistry.json
```diff
@@ -4078,6 +4078,24 @@
...
+        "publications": [
+          {
+            "doi": "10.1093/database/baae123",
+            "pmc": "PMC11651143",
+            "pubmed": "39689042",
+            "title": "AthRiboNC: an Arabidopsis database for ncRNAs with coding potential revealed from ribosome profiling",
+            "year": 2024
+          }
+        ],
+        "uri_format": "https://bis.zju.edu.cn/athribonc/miRNA/AT5G07875"
```
bgyori, line 4096:
```suggestion
        "uri_format": "https://bis.zju.edu.cn/athribonc/miRNA/$1"
```

bgyori:
I realize this is actually problematic because technically the AT5G07875 is a TAIR locus name, not an identifier (called "TAIR Accession" in their terminology), and the current `tair.locus` (similar to other subprefixes for `tair`) is defined for identifiers like 6532565372. There is generally a complicated situation surrounding TAIR identifiers and their resolution, as noted in #1335, so whether we want to add this or not requires a bit more thought.

bgyori:
Given the overall situation (see previous comments), I think we can omit this resource so I'll close this

PR #1386, opened 2025-01-28
Update URI formats for tair.locus and tair.protein

bgyori:
This PR partially addresses #1335 by adding new URI formats for `tair.locus` and `tair.protein`. The latter is using a pattern that @nagutm found, and after some trial and error, I found a URI format that works for the former. I didn't manage to find any working URI format for resolving `tair.gene` identifiers.


cthoyt (changes requested):
Please make sure not to throw away old URI formats when making updates, these are still required to parse data that is out in the wild. When making these kinds of updates, one can put the old URI formats into a provider, where one can also give context as to why they were removed in the description field

bgyori:
I agree preserving the old URL has some utility but if we curate it as a provider, wouldn't it appear as a broken resolver link on the website? Would it make sense to introduce a dedicated slot in the schema for legacy URLs that are known to be outdated but are used specifically for the purpose of reverse mapping of URLs to CURIEs?

cthoyt:
Agreed, I think introducing a controlled vocabulary for "status" in the Provider class would accomplish this - then we could flag things as being "legacy" URLs that either still can be resolved or not.

bgyori:
I added a legacy provider following the pattern used in #1411. When https://github.com/biopragmatics/bioregistry/issues/1387 is implemented, we can go back and make these legacy status flags more formal.

Issue #1383, opened 2025-01-28
Update UniProt Protein


JervenBolleman:
### Prefix

uniprot

### Explanation

Please update the contact details to be current (Rolf has not been working on UniProt  for years) use www.uniprot.org/contact instead.

### Contributor ORCID

0000-0002-7449-1266

cthoyt:
Hi @JervenBolleman, can you propose an alternative primary contact for the `uniprot` prefix?

Note that similarly to OBO Foundry, it's [Bioregistry policy](https://github.com/biopragmatics/bioregistry/blob/main/docs/CONTRIBUTING.md#contact-and-attribution) to curate a primary responsible individual, instead of contact lists.

JervenBolleman:
@cthoyt I can't, as it is UniProt policy that all requests are send to the help desk. 

If you need people then it is
https://www.uniprot.org/help/bateman
https://www.uniprot.org/help/bridge
and
https://bioinformatics.udel.edu/people/cathy-wu-phd/

You should add, on all uniprot related records, and remove any other e-mail addresses that might be there.
I strongly suggest you do use the help desk as a contact method, because to many user-requests have been missed over the years by trying to reach certain PIs instead of help-desks.

cthoyt:
@JervenBolleman is there a direct contact email address?

JervenBolleman:
@cthoyt no, the contact form is the system to use. This is common across the large data set providers and institutes. 

I appreciate that you opened https://github.com/biopragmatics/bioregistry/issues/1385 and hope to contribute the why there.

cthoyt:
thanks @JervenBolleman. I also opened https://github.com/biopragmatics/bioregistry/pull/1392 to make a more explicit place to store the contact page, rather than trying to conflate it with other fields that are better suited for keeping track of email addresses.

PR #1392, opened 2025-01-30
New schema field - contact page

cthoyt:
Motivated by #1383, this PR adds a new schema element for web-based contact forms. In order to promote transparency, we'll still require a primary contact person to be set in order to use this field.

- [x] add new schema element
- [x] add example of using schema element (uniprot)
- [x] add component to web UI
- [x] add curation guide
- [x] add unit tests for data integrity


bgyori:
This one seems complete to me, should we try to merge it or are there any further todos?

PR #1393, opened 2025-01-31
New prefix: mirgenedb

nalikapalayoor:
This PR curates a new prefix microRNA Gene Database. 

https://pubmed.ncbi.nlm.nih.gov/39673268/


src/bioregistry/data/bioregistry.json
```diff
@@ -70498,6 +70498,35 @@
...
+      "github": "nalikapalayoor",
+      "name": "Nalika Palayoor",
+      "orcid": "0009-0008-8406-631X"
+    },
+    "description": "mircoRNA Gene Database is a collection of animal microRNA genes. This database allows researchers to analyze mircoRNA properties like secondary structure and features in a phylogenetic context.",
+    "example": "hsa/Let-7-P1b",
+    "github_request_issue": 1393,
+    "homepage": "https://mirgenedb.org/",
+    "name": "microRNA Gene Database",
+    "pattern": "^[a-zA-Z0-9-]+/[a-zA-Z0-9-]+$",
```
bgyori, line 70518: Somewhat inconveniently from the perspective of the `uri_format`, it looks like the actual IDs use `-` rather than `/`. We should prioritize curating the ID accurately so I would suggest using `-` here. Then, the `uri_format` could be updated to something like https://mirgenedb.org/browse?qtype=mgid&org=ALL&query=$1 which is a search endpoint but correctly resolves these IDs.

image

src/bioregistry/data/bioregistry.json
```diff
@@ -70498,6 +70498,35 @@
...
+    "publications": [
+      {
+        "doi": "10.1093/nar/gkae1094",
+        "pmc": "PMC11701709",
+        "pubmed": "39673268",
+        "title": "MirGeneDB 3.0: improved taxonomic sampling, uniform nomenclature of novel conserved microRNA families and updated covariance models",
+        "year": 2025
+      }
+    ],
+    "uri_format": "https://mirgenedb.org/show/$1"
```
bgyori, line 70528: There are multiple prior publications about MirGeneDB that we should curate here:

 MirGeneDB is a database of manually curated microRNA genes that have been validated and annotated as initially described in [Fromm et al. 2015 ](http://www.annualreviews.org/doi/abs/10.1146/annurev-genet-120213-092023), [Fromm et al. 2020](https://academic.oup.com/nar/advance-article/doi/10.1093/nar/gkz885/5584683?guestAccessKey=b9fe4339-a36b-49df-96ea-9e2cd7a1c99b) and [Fromm et al 2022](https://academic.oup.com/nar/advance-article/doi/10.1093/nar/gkab1101/6439665). MirGeneDB 3.0 ([Clarke and Hoye et al. 2024 ](https://academic.oup.com/nar/advance-article/doi/10.1093/nar/gkae1094/7914206?login=true))

src/bioregistry/data/bioregistry.json
```diff
@@ -70498,6 +70498,56 @@
...
+      "orcid": "0000-0003-0352-3037"
+    },
+    "contributor": {
+      "email": "palayoor.n@northeastern.edu",
+      "github": "nalikapalayoor",
+      "name": "Nalika Palayoor",
+      "orcid": "0009-0008-8406-631X"
+    },
+    "description": "mircoRNA Gene Database is a collection of animal microRNA genes. This database allows researchers to analyze mircoRNA properties like secondary structure and features in a phylogenetic context.",
+    "example": "hsa/Let-7-P1b",
```
bgyori, line 70514: This is not a valid ID, it could be e.g., Hsa-Let-7-P2a3 as in my previous comment

src/bioregistry/data/bioregistry.json
```diff
@@ -70498,6 +70498,56 @@
...
+    "github_request_issue": 1393,
+    "homepage": "https://mirgenedb.org/",
+    "name": "microRNA Gene Database",
+    "pattern": "^[a-zA-Z0-9-]+-[a-zA-Z0-9-]+$",
```
bgyori, line 70518: This pattern doesn't entirely make sense because the `-` is matched as a character in the first `[a-zA-Z0-9-]` block as many times as needed, so the whole second part of the pattern `-[a-zA-Z0-9-]+` is redundant. It also doesn't exclude an identifier starting with a `-`, etc. It should be something like `^[a-zA-Z0-9]+(-[a-zA-Z0-9]+)+$`

src/bioregistry/data/curated_papers.tsv
```diff
@@ -1,4 +1,7 @@
 pubmed	relevant	orcid	date_curated	relevancy_type	pr_added	notes
+26473382	1	0009-0008-8406-631X	2025-2-1	new_publication	1394	Publication for mirgenedb
+31598695	1	0009-0008-8406-631X	2025-2-1	new_publication	1394	Publication for mirgenedb
+34850127	1	0009-0008-8406-631X	2025-2-1	new_publication	1394	Publication for mirgenedb
```
bgyori, line 4: The three old papers that didn't come out of the prediction pipeline can be removed from this file

PR #1398, opened 2025-02-01
New prefix: psskb

nalikapalayoor:
PSSKB uses identifiers for protein motifs for which it provides data
Pubmed: https://pubmed.ncbi.nlm.nih.gov/39876062/
Database: https://psskb.org/


src/bioregistry/data/bioregistry.json
```diff
@@ -88335,6 +88335,21 @@
...
+        "name": "Protein Structure and Stability Knowledge Base",
+        "publications": [
+          {
+            "doi": "10.1002/jcc.70046",
+            "pubmed": "39876062",
+            "title": "PSSKB: A Web Application to Study Protein Structures",
+            "year": 2025
+          }
+        ],
+        "uri_format": "https://www.rcsb.org/structure/$1"
```
bgyori, line 88352: The `uri_format` here would have to be a PSSKB URL, not a PDB URL. Same for UniProt below.

bgyori:
Upon closer review, I am not convinced this should be a provider for UniProt and PDB. All the UniProt/PDB identifiers just link to those external resources, PSSKB doesn't actually appear to resolve these. What PSSKB provides is entries for what they call Motifs such as at https://psskb.org/motif/card/2VM6_A55_A135 (see also screenshot). 

image

For practical purposes the Motif IDs could be considered unique IDs introduced and resolved by PSSKB (even though they are not completely novel in the they consist of three pieces, a PDB ID, a start marker and an end marker). Some examples here:

image

If we do this then it could be considered a new prefix.

src/bioregistry/data/bioregistry.json
```diff
@@ -88336,6 +88364,9 @@
         ],
         "uri_format": "https://seq2fun.dcmb.med.umich.edu/furna/pdb.cgi?pdbid=$1"
       }
+        ],
+        "uri_format": "https://www.rcsb.org/structure/$1"
+      }
```
bgyori, line 88369: This change should be removed, right?

nalikapalayoor:
There is no contact info listed anywhere on the website for the database so I put placeholders for now (so that I could lint the file). What should I do about this?

src/bioregistry/data/bioregistry.json
```diff
@@ -96364,6 +96364,34 @@
     "homepage": "https://wiki.phenoscape.org/wiki/Curation_workflow",
     "name": "Phenoscape Publication"
   },
+  "psskb": {
+    "contact": {
+      "email": "not_listed",
+      "name": "not listed",
+      "orcid": "0000-0000-0000-0000"
```
bgyori, line 96371: Generally, if some information is missing, we should not add "filler" data like this, we would just leave it out. Please use the corresponding author's details available here: https://onlinelibrary.wiley.com/doi/10.1002/jcc.70046

nalikapalayoor: Ok that makes sense. I found his name and orcid but was not able to find his email. I put his company email down, do you think that makes sense or to just leave it out?

cthoyt:
@nalikapalayoor this curation training video might be useful for you for finding contact info in the future https://www.youtube.com/watch?v=FN6JJ0oKNCU

src/bioregistry/data/curated_papers.tsv
```diff
@@ -169,3 +169,4 @@ pubmed	relevant	orcid	date_curated	relevancy_type	pr_added	notes
 39728342	0	0009-0008-8406-631X	2025-01-17	non_resource_paper	1364	
 39729356	0	0009-0008-8406-631X	2025-01-14	not_identifiers_resource	1364	
 39732706	0	0009-0008-8406-631X	2025-01-10	not_identifiers_resource	1351	
+39876062	1	0009-0008-8406-631x	2025-02-01	new_prefix	1398
```
bgyori, line 172:
```suggestion
39876062	1	0009-0008-8406-631X	2025-02-01	new_prefix	1398	
```

nalikapalayoor:
@cthoyt I will look at that thank you so much!

bgyori:
Something got messed up with this git branch... I can help offline.

bgyori (approved):

PR #1402, opened 2025-02-02
Adjust the dates for paper_ranking pipeline to be from the first to the last of the previous month

nagutm:
This pull request modifies the `start_date` and `end_date` variables of the `paper_ranking` pipeline to be set to the first and last date of the month respectively. 

The old variables were simply subtracting 30 days from the start date which was causing it to miss out on ranking papers from the 1st of the month in some cases. 

I also modified the `cron` job to run at midnight EST instead of midnight UTC which is what it was set to previously.

@cthoyt

### DIFF
--- a/.github/workflows/paper_ranking.yml
+++ b/.github/workflows/paper_ranking.yml
@@ -2,7 +2,7 @@ name: Run Paper Ranking Script and Update Issue
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


bgyori (approved):
This looks good to me, I tested the date logic on a few examples and it seems to work as expected.

Issue #1410, opened 2025-02-10
Tidying up the uniprot namespaces


cmungall:
If you look at the uniprot site:

https://www.uniprot.org/

We see there are clear divisions in the uniprot *resource* into different areas. The 4 primary ones (proteins, species/proteomes, clusters, uniparc), and additional ones (e.g keywords).

Of the 4 primary, of course "proteins" are the focal entry, and this is referred to as `UniProtKB`, as can be seen in the URLs, e.g

https://www.uniprot.org/uniprotkb/P05067/entry

Unfortunately this clean hierarchical subdivision was completely messed up by identifiers.org, and inherited by bioregistry, so we have:

* https://bioregistry.io/registry/uniprot (actually uniprotkb)
* https://bioregistry.io/registry/uniprot.proteome
* https://bioregistry.io/registry/uniref
* https://bioregistry.io/registry/uniparc
* https://bioregistry.io/registry/uniprot.keyword
* ...

There is no rhyme or reason as to what gets a lot level namespace vs what gets a sub-namespace.

Adding to the confusion, the GO registry and its offshoots has always used the UniProt-blessed nomenclature, and we use `UniProtKB` as a prefix. We actually used to use `UniProt`, but changed it to `UniProtKB`. However, this wasn't universally taken up by the community, we can see that many resources continue to use `UniProt` as the prefix when referring to UniProtKB entries, even ones within EBI, e.g.

https://www.ebi.ac.uk/interpro/protein/UniProt/P05067/entry/pfam/#table

And the PURLs use uniprot:

http://purl.uniprot.org/uniprot/Q6NVU6

This is not unreasonable, even though strictly speaking P05067 is in UniProtKB, this is the focal database, and let's face it, nobody uses the term "uniprotkb:

<img width="1160" alt="Image" src="https://github.com/user-attachments/assets/9c45da84-a84a-4206-8945-d328c0dd5f75" />

This is in contrast to

- #1404

Where the bioregistry/id.org entry makes no sense.

We would like to bring GO in line with the rest of the community, and we are exploring what it would cost to revert back from `UniProtKB` to `UniProt`, so I am opening up this issue for broader community discussion, in particular from the UniProt (sensu broad) team.

[...]

PR #1411, opened 2025-02-11
Fixing GO entries

cmungall:
- Fixing URL for GO CAMs (go.model)
- Fixed error in name (Activity not Assembly)
- Fixed URL for GO_REFs


cthoyt:
thanks @cmungall. I updated your PR to retain the legacy URLs, since they probably appear in sources we want the Bioregistry to support parsing the URLs from.

I also added a few more flags for other GO prefixes, perhaps you can do an audit and make a follow-up PR to clarify who's the contact for each GO prefix, what the canonical prefix should be (similar to your request in #1410), and anything else that can improve those records

cmungall:
The json looks good but I'm a bit confused as to why this hasn't propagated to the site yet...

cthoyt:
> The json looks good but I'm a bit confused as to why this hasn't propagated to the site yet...

Sorry about this, it's related to how the bioregistry's CI/CD works. I opened a new issue for fixing this in https://github.com/biopragmatics/bioregistry/issues/1416

Issue #1404, opened 2025-02-05
EC should be the primary prefix for EC, not eccode


cmungall:
Everyone used `EC` until id.org unilaterally decided it was `eccode`, which nobody used. Unfortunately bioregistry inherited this from id.org and did not use the GO registry which reflected actual use

We should reverse this decision, specifically

1. Make `EC` primary
2. Make the preferred casing be `EC` not `ec`
3. Keep `eccode` as a deprecated synonym

Evidence for `EC` being more widespread:

 * https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=%22eccode%3A3.1.1.1%22&btnG=. **0 results**
 * https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=%22EC%3A3.1.1.1%22&btnG= **8k results** albeit most without an explicit `:`

The following databases all use `EC`

* PubChem: https://pubchem.ncbi.nlm.nih.gov/protein/EC:3.1.1.1
* GO
* Alliance of Genome Resources
* Expasy https://enzyme.expasy.org/EC/3.1.1.1 
* KEGG https://www.genome.jp/dbget-bin/www_bget?ec:3.1.1.1
* RHEA https://www.rhea-db.org/rhea/?query=EC:2.2.1.6
* biolink
* any everyone else

(I am not sure I have time to do make tickets for all instances of this but EC is one of the more egregious ones. I could make a mega-ticket if it helps)

cthoyt:
I have no opinion on this either way. Compared to the dashes discussion, I don’t think there is a ton of community disagreement on this, and there’s a lower impact on anyone who is using Bioregistry code to keep their data standardized. I’d give it a plus 1

bgyori:
I think this is a good idea, at least in principle. If we proceed with this, a few tests, benchmarks and documentation sections will also have to be updated where EC/eccode is referenced: https://github.com/search?q=repo%3Abiopragmatics%2Fbioregistry%20eccode&type=code. It also appears in various test cases in other repos like https://github.com/search?q=eccode%20repo%3Abiopragmatics%2Fpyobo%20&type=code where the current Bioregistry standard prefix is assumed.

cthoyt:
Closed in #1413, see https://bioregistry.io/ec.

@cmungall please feel free to propose as many specific issues as you'd like for other prefix cleanup. This would be nicer and easier to track progress than a mega-issue (though we could also have an "epic" issue to group them if this is helpful for discussion)

PR #1413, opened 2025-02-12
Rename eccode to ec

nagutm:
This pull request addresses the following issue: https://github.com/biopragmatics/bioregistry/issues/1404

The following changes are implemented:
- Rename `eccode` prefix to `ec`
- `preferred_prefix` is set to `EC`
- Maintain `eccode` as a synonym
- Update any instances of `provides: eccode` to  `provides: ec`

A few additional files that referenced `eccode` are updated to reflect this change as well.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -70,7 +70,7 @@ that's required to go with a given prefix.
 Unfortunately, these requirements can not be applied retroactively and can not
 be trivially applied to automatically imported prefixes. In some cases,
 historical prefixes can be modified to follow these requirements. For example,
-Identifiers.org's `ec-code` was renamed to `eccode` while maintaining `ec-code`
+Identifiers.org's `ec-code` was renamed to `ec` while maintaining `ec-code`
 as a synonym.
 
 Original discussion about minimum prefix requirements can be found at
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -70,8 +70,8 @@ that's required to go with a given prefix.
 Unfortunately, these requirements can not be applied retroactively and can not
 be trivially applied to automatically imported prefixes. In some cases,
 historical prefixes can be modified to follow these requirements. For example,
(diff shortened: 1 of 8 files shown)


cthoyt (approved):

bgyori (approved):
Looks great, thanks!

bgyori:
Oh actually, just noticed one more small thing, we might want to update this sentence: https://github.com/biopragmatics/bioregistry/blob/7fae7ba39716b09c6bde4c42e4dc422ade79b88d/docs/CONTRIBUTING.md?plain=1#L73

cmungall (approved):

docs/CONTRIBUTING.md
```diff
@@ -70,7 +70,7 @@ that's required to go with a given prefix.
 Unfortunately, these requirements can not be applied retroactively and can not
 be trivially applied to automatically imported prefixes. In some cases,
 historical prefixes can be modified to follow these requirements. For example,
-Identifiers.org's `ec-code` was renamed to `eccode` while maintaining `ec-code`
+Identifiers.org's `ec-code` was renamed to `ec` while maintaining `ec-code`
```
cthoyt, line 73: this is a bit misleading. The point of this paragraph is to explain the normalization that happens to external prefix registries' prefixes. Maybe we need to find a different example from identifeirs.org

cthoyt: I just ran the following script and didn't find any other good examples to illustrate prefix normalization, so I think it's okay to just leave it since ec-code was the only good example here and there isn't a more general thing going on

```python
import bioregistry

if __name__ == '__main__':
    for resource in bioregistry.resources():
        miriam = resource.get_miriam_prefix()
        if not miriam or not "-" in miriam:
            continue
        if miriam.replace("-", "") == resource.prefix:
            print(resource.prefix)
```

bgyori: Yes, this change makes the example a bit awkward but it is probably okay to keep

PR #1426, opened 2025-02-18
New publication: encode

nalikapalayoor:
This PR adds a new publication for the "encode" prefix:
https://pubmed.ncbi.nlm.nih.gov/39763870/


bgyori:
It looks like this is a preprint and it might not be worth curating preprints as publications for existing prefixes, since these preprints are likely to be published later and their PMID/PMCID/DOI are subject to change, which we would then have to update.

PR #1429, opened 2025-02-21
Allow provider specific examples

### LINKED ISSUE #1417: Allow provider-specific example ID
Should we extend the provider data model with an **optional provider-specific example ID** for the purpose of having working links in the provider table? There are other examples of this issue, for instance, for https://bioregistry.io/registry/uniprot, we have 
image
as a provider but https://www.iedb.org/antigen/UNIPROT:P0DP23 isn't an entry in IEDB, however, some other UniProt ID that has an actual entry would work well as an example link.

_Originally posted by @bgyori in https://github.com/biopragmatics/bioregistry/issues/1282#issuecomment-2659478039_

nagutm:
Closes #1417

It updates the web front-end to check if a provider-specific example is available and, if so, modifies the provider link in the Extra Providers sections to use the custom example ID instead of the default one.

As a test I added a custom example for `bfvd` which is a provider for `uniprot` and it correctly displays the link with the custom example and is able to resolve the ID appropriately.

image


cthoyt (changes requested):
@nagutm thanks for looking into this! One thing that's important for any new schema change proposals is to add a real curation to bioregistry.json that reflects usage of the new schema element.

Specifically for this one, we want to make sure we apply the regular expression pattern to validate the ID is valid. You can use the code that checks the primary example / example extras as reference

Speaking of example extras, there should be a nice way of aggregating any possible specific examples put in the providers into the example_extras field. The question is: should we require double curation there, or should we have the function that wraps it automatically grab and aggregate them? Check usages of `.example_extras` to see what might be more comfy

nagutm:
> Specifically for this one, we want to make sure we apply the regular expression pattern to validate the ID is valid. You can use the code that checks the primary example / example extras as reference

@cthoyt Could you clarify exactly what piece of code you are referring to here that checks the primary example/example extras?

nagutm:
> Speaking of example extras, there should be a nice way of aggregating any possible specific examples put in the providers into the example_extras field. The question is: should we require double curation there, or should we have the function that wraps it automatically grab and aggregate them? Check usages of `.example_extras` to see what might be more comfy

I added a method to the Resource class called `get_aggregated_example_extras()` which concatenates the manually curated example_extras along with any provider specific examples. I also updated any references to `.example_extras` with this method to ensure that provider-specific examples are included in any list of example_extras in the future.

bgyori:
> > Specifically for this one, we want to make sure we apply the regular expression pattern to validate the ID is valid. You can use the code that checks the primary example / example extras as reference
> 
> @cthoyt Could you clarify exactly what piece of code you are referring to here that checks the primary example/example extras?

There are tests that check whether example identifier for a given entry conform to the identifier regular expression pattern definition. These tests should cover not only the default and extra examples but also the examples specific to providers so if an invalid example identifier is added, the test fails.

tests/test_data.py
nagutm, line 1: I believe that by replacing `.example_extras` with `.get_aggregated_example_extras` this will assert that any provider specific examples are also conforming to the identifier pattern for the resource.

nagutm:
I added a provider example for `furna` to run some of the tests implemented in this PR. The previous link: https://seq2fun.dcmb.med.umich.edu/furna/pdb.cgi?pdbid=2gc4 was not resolving the default `pdb` ID as expected.

tests/test_data.py
```diff
@@ -744,6 +744,13 @@ def test_providers(self):
                             f"in prefix publication list (appears as {other.title})",
                         )
                         self.assert_publication_identifiers(publication)
+                    # Check provider-specific example handling
+                    if provider.example:
```
cthoyt, line 748: Is there not already a test that checks there's a `$1` character in the provider's URI format?

I think this added block can be removed

src/bioregistry/schema/struct.py
```diff
@@ -1353,8 +1361,21 @@ def get_examples(self) -> list[str]:
         example = self.get_example()
         if example:
             rv.append(example)
-        rv.extend(self.example_extras or [])
+        rv.extend(self.get_aggregated_example_extras() or [])
         return rv
+    
+    def get_aggregated_example_extras(self) -> list[str]:
```
cthoyt, line 1367: Please rename to `get_example_extras`

src/bioregistry/schema/struct.py
```diff
@@ -1353,8 +1361,21 @@ def get_examples(self) -> list[str]:
...
+        Combine manually curated examples with provider-specific examples.
+        """
+        aggregated_examples = set(self.example_extras or [])
+
+        if self.providers:
+            for provider in self.providers:
+                if provider.example:
+                    aggregated_examples.add(provider.example)
+
+        return list(aggregated_examples)
```
cthoyt, line 1378: for consistency

```suggestion
        return sorted(aggregated_examples)
```

cthoyt (changes requested):
besides the two comments I left, this is ready to go. Thanks @nagutm

cthoyt (approved):

PR #1436, opened 2025-02-25
Exclude preprints from ranking and improve metadata access

bgyori:
This PR implements an exclusion list for journals to be used in the automated paper ranking. I added bioRxiv and medRxiv to this list but in principle others can be added if necessary. The PR also simplifies the way we get metadata using a function that upstreams the issue of handling a large number of PMIDs using a function that has been around in INDRA for a while but for some reason not used in this script.

Background: Recently, PubMed is assigning PMIDs to preprints on bioRxiv and medRxiv and these have been showing up in the paper ranking. It is generally not a good idea to curate these as publications associated with Bioregistry entries since any such curation would have to be updated when the paper is published with its stable metadata (title, DOI, PMID, etc. can all change upon publication) which is difficult to coordinate since time of actual publication is hard to monitor. See e.g., #1425 and #1426.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -141,7 +141,6 @@ paper-ranking = [
     "pandas",
     "scikit-learn",
     "tabulate",
-    "more_itertools",
 ]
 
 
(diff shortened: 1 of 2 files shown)


bgyori:
Tested locally and this works as expected, preprints are removed from predictions.

PR #1440, opened 2025-02-26
Improve support for federated queries to the SPARQL endpoint

vemonet:
Upgrade rdflib-endpoint to a version where federated queries from ontotext GraphDB and blazegraph are fixed @cthoyt


cthoyt (approved):
thanks @vemonet !

PR #1450, opened 2025-03-07
Add mechanism for filtering non-W3C-compliant CURIE prefix synonyms

### LINKED ISSUE #1338: Prefix with space character?
Check this: https://github.com/mapping-commons/sssom-py/pull/568/files#r1886866051

It contains a prefix with a space:

```
"prefix_synonyms": [
            "MEDLINE",
            "PubMed",
            "PubMed ID",
```

For various reasons, I think it makes sense to not allow this. For example:

https://github.com/mapping-commons/sssom-py/pull/568/files#r1886866270

```
http://bioregistry.io/PubMed ID:
```

Which really doesn't make any sense at all. 

I think we should add qc that prevents spaces in prefixes.

cthoyt:
Closes #1338

This PR adds `enforce_w3c` flag to a Bioregistry context, which when turned on, filters out CURIE prefix synoynms that don't match the W3C regex for NCNAME.

@matentzn this PR also turns this flag on for the OBO context. Can you review that you're happy with the results?


matentzn (approved):
YES this is amazing!!

PR #1451, opened 2025-03-07
Add descriptions for external TogoID entries

bgyori:
This PR implements retrieving descriptions for external TodoID entries and adding them to the pre-processed data artifact. This is beneficial for workflows such as #1439 to which additional description serves as important context in understanding what exact resource a prefix represents.

### DIFF
--- a/src/bioregistry/external/togoid/processed.json
+++ b/src/bioregistry/external/togoid/processed.json
@@ -85,6 +85,7 @@
   },
   "Bioproject": {
     "catalog": "nbdc00476",
+    "description": "BioProject is a database which collects the results of research projects under the umbrella of the project or initiative. The database consists of tables of projects, which users can search or browse to find all data outputs related to that project. Users can submit projects or download data via FTP.",
     "examples": [
       "PRJDB575",
       "PRJDB1880",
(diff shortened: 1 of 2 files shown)


bgyori:
@cthoyt do you see any issues with this? If not, I will build some new things on top of it.

src/bioregistry/external/togoid/processed.json
```diff
@@ -195,6 +198,7 @@
   },
   "Chebi": {
     "catalog": "nbdc00027",
+    "description": "ChEBI is a collection of data on entities of biological interest.\r\nThis database concentrates on assembling data on smaller chemical compounds, such as atoms, molecules, ions, radicals etc. These should be produced in nature or be synthetic molecules affecting biological processes in living organisms.\r\nChEBI does not generally include molecules directly encoded by the genome.\r\nThe database contains 31,651 entries with annotations.\r\nA keyword search option is available.\r\nIn addition to this an advanced search option, which includes a structure drawing tool, is provided.\r\nThe ChEBI database can be searched individually or in combination with other databases.",
```
cthoyt, line 201: Not clear if keeping in newlines here is good or can be problematic, but at least we should normalize the `\r` carriage returns out of it.

cthoyt: otherwise, all good! thanks!

bgyori: Okay, I removed all of those, so we just get a space instead of all new lines / carriage returns which I think should be fine.

PR #1453, opened 2025-03-08
New provider: gwh

nalikapalayoor:
The corona section of the Genome Warehouse uses Genbase to resolve identifiers
https://ngdc.cncb.ac.cn/gwh/browse/virus/coronaviridae


bgyori:
My take is that this is not useful as a provider: GWH here doesn't provide its own data for GenBank identifiers, it just lists GenBank identifiers that link to GenBank itself. So I will close this for now.

PR #1454, opened 2025-03-09
Add prefix: phagedive

nagutm:
This PR curates a new prefix: `phagedive`.

PubMed: https://pubmed.ncbi.nlm.nih.gov/39373542/
Database: https://phagedive.dsmz.de/

It is a database of bacteriophages and archaeal viruses and provides details regarding their taxonomy, host strain, phage morphology, life cycle, origin and genomic data.

### DIFF
--- a/src/bioregistry/data/curated_papers.tsv
+++ b/src/bioregistry/data/curated_papers.tsv
@@ -58,6 +58,7 @@ pubmed	relevant	orcid	date_curated	relevancy_type	pr_added	notes
 39341994	0	0009-0009-5240-7463	2024-10-15	not_identifiers_resource	1223	
 39345624	1	0009-0009-5240-7463	2024-10-19	new_publication	1223	Publication for sgd and sgd.pathways
 39373530	1	0009-0009-5240-7463	2025-02-15	new_prefix	1418	
+39373542	1	0009-0009-5240-7463	2025-03-09	new_prefix	1454	
 39379619	1	0009-0009-5240-7463	2024-11-26	new_provider	1285	Provider for UniProt
 39399372	0	0009-0009-5240-7463	2024-11-19	irrelevant_other	1288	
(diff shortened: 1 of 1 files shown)


cthoyt:
we should consider the relation to DSMZ's other resource in the bioregistry "CellDive" https://bioregistry.io/registry/dsmz, which has a top-level prefix based on the organization itself. Maybe we subspace both under dsmz.celldive and dsmz.phagedive

nagutm:
In addition to `CellDive`, there is also `MediaDive` and `BacDive` which also fall under the DSMZ umbrella. Since these other prefixes are not subspaced under dsmz, could we instead consider renaming the current `dsmz` prefix to `celldive` which is more specific and update the metadata to reflect this change? 

Additionally, the `cellosaurus` mapping for the current `dsmz` prefix is also for CellDive, not DSMZ as a whole. The `re3data`, however, is for DSMZ as a whole and would probably need to be removed.

cthoyt:
@nagutm that's a good point! I think there are a lot of references to `dsmz` because the cell line resource was their first big informatics one, but I agree that swapping this out for celldive would do good in the long run

src/bioregistry/data/bioregistry.json
```diff
@@ -14087,6 +14087,34 @@
     },
     "uri_format": "https://www.cellbankaustralia.com/$1.html"
   },
+  "celldive": {
```
cthoyt, line 14090: changing the prefix for celldive does not belong in this PR

nagutm: I'll make a separate one

cthoyt: thanks!

nagutm: See https://github.com/biopragmatics/bioregistry/pull/1478

bgyori (approved):

PR #1455, opened 2025-03-09
Add provider: sabdab

nagutm:
This PR curates a new provider `sabdab` (The Structural Antibody Database) for `pdb`.

It is also curates a custom example, since the default ID (2gc4) cannot be resolved by this provider.


bgyori (approved):
This looks good, thanks!

PR #1457, opened 2025-03-10
Update curated mappings with mismatches

bgyori:
This PR adds all the mapping mismatches I found while reviewing some mappings (#1456).


bgyori:
I'm tempted to close this one since this would be better integrated on top of #1458

bgyori:
If we want to be really pedantic about this, technically `semapv:MappingReview` (A process that is concerned with determining if a mapping “candidate” (otherwise determined) is reasonable/correct.) might be a better predicate when using the automated scoring table as a basis for curation than `semapv: ManualMappingCuration` (A matching process that is performed by a human agent and is based on human judgement and domain knowledge.) so I'll update these.

PR #1458, opened 2025-03-10
Replace mismatch file with more general mappings table

bgyori:
**Background**: After #1439 and #1456 it makes sense to systematically review certain mappings and record incorrect ones (currenty in mismatch.json) so they can be removed. However, as part of review, one naturally encounters correct mappings as well which need to be recorded somewhere to enable benchmarking and to avoid re-review of the same mappings redundantly. A further shortcoming of the current mismatch.json is that it is structurally limited to representing a single prefix per external registry.

**Proposal**: This PR adds a new TSV file at data/curated_mappings.tsv which serves as a place to add the result of mapping review. I propagated all the existing mismatches from mismatch.json into curated_mappings.tsv. The PR also adds a documentation guide to help with the mapping review process.

**Questions**:
- If we have a dedicated file for mapping review, what should we do with mismatch.json?
   - Option 1: just remove mismatch.json and use curated_mappings.tsv instead as the single source of truth. I don't think there is any downstream usage of mismatch.json so making this change shouldn't be disruptive. Internal code related to the use of mismatch.json would have to be updated.
   - Option 2: Preserve mismatch.json but only as output of an automated workflow that generates it from the TSV.
   - Option 3: The two files could be manually kept in sync which requires no further coding but is not great otherwise.
- Should curated_mappings.tsv actually be a TSV similar to curated_papers.tsv or should it be JSON similar to mismatch.json just with an enhanced data model?
- I intentionally described this as mapping "review" rather than mapping "curation" since the idea is to review mappings that _already exist_. Would it make sense to use this same file to curate _additional_ mappings that were not otherwise produced and use these during alignment?

**TODOs**:
- Add validity checking for curated_mappings.tsv similar to those for curated_papers.tsv
- Extend the curation guide: further guidance and a specific example


cthoyt:
I think this is a good idea - given these are mappings, let's use the SSSOM schema for the TSV file since it already has well-defined fields for the Bioregistry use case (justification based on manual curation or automated, provenance for contributor, negative assertions).

mismatch.json and any manually curated positive mappings could go into this SSSOM file.

Later, it might make sense to incorporate predictions from the text matching workflow, or even split the mappings out of the bioregistry.json file. I've actually considered doing this before, but put it off since it would cause a lot of diff and would require re-investigating how metadata gets propagated. Further, the existing alignment workflow has relatively high precision, so this never became a priority. It's also worth noting the mappings in bioregistry.json are a mix of automatic and manual assertions, so it wouldn't be easy to figure out where any given mapping came from.

bgyori:
Alright, so we can change the file to SSSOM format and then update the internal code to use it instead of mismatch.json. For this to be self contained, ideally we should do all of that on this PR.

bgyori:
Added initial SSSOM implementation of mappings, anything else to include/change?

src/bioregistry/data/curated_mappings.tsv
```diff
@@ -0,0 +1,101 @@
+subject_id	predicate_modifier	predicate_id	object_id	creator_id	comment
```
cthoyt, line 1: for this to be valid SSSOM, it needs a [`mapping_justification`](https://mapping-commons.github.io/sssom/mapping_justification/) column. For all of these entries so far, the right CURIE is `semapv:ManualMappingCuration`. Later, we can also use other terms in `semapv`

we also will want to have a new set of unit tests to test the integrity of the format and content in this file.

bgyori:
@cthoyt should we merge this or is there anything else to consider? We have a bunch of curations ready that can be added on top of this in separate PRs once it's merged.

cthoyt (approved):
I renamed the curation file to have .sssom.tsv extension and ran ruff to get rid of unnecessary typing imports. I left an optional comment about switching to using pydantic models to reduce complexity.

Other than that, this is ready to merge

docs/guides/mapping_review.md
```diff
@@ -0,0 +1,43 @@
...
+harmonizes across entries of multiple external registries via **mappings**. Each
+mapping is between a Bioregistry prefix and a specific prefix within one of the
+external registries the Bioregistry integrtes with. The initial creation of such
+mappings is largely automated. The vast majority of these mappings is correct
+but some correspond to prefixes that do not represent the same resource (or more
+specifically, the same semantic space) and should therefore not be mapped to
+each other.
+
+The following
+[TSV file](https://github.com/biopragmatics/bioregistry/blob/main/src/bioregistry/data/curated_mappings.tsv)
```
cthoyt, line 17: mention this is SSSOM, link to external documentation on format and usage

docs/guides/mapping_review.md
```diff
@@ -0,0 +1,43 @@
...
+- `subject_id`: The CURIE corresponding to Bioregistry prefix from which this is
+  a mapping. Even when a mapping is symmetric, as a convention, this should
+  always be the Bioregistry prefix not the external one.
+- `predicate_modifier`: A modifier on the predicate, typically `Not` if the
+  mapping is negative.
+- `predicate_id`: The CURIE representing the predicate of the mapping, for
+  instance `skos:exactMatch`.
+- `object_id`: The CURIE corresponding to the external prefix to which this is a
+  mapping.
+- `curator_id`: The ORCID of the curator reviewing the mapping represented as a
```
cthoyt, line 30:
```suggestion
- `creator_id`: The ORCID of the curator reviewing the mapping represented as a
```

PR #1478, opened 2025-03-22
Add prefix: celldive, separating it out from dsmz

nagutm:
Based on the discussion in https://github.com/biopragmatics/bioregistry/pull/1454, I renamed the old `dsmz` prefix which was too vague to `celldive` which better reflects the specific semantic space that the metadata was describing.

I updated the description, mappings, regex pattern and examples to reflect this change.


src/bioregistry/data/bioregistry.json
```diff
@@ -14087,6 +14087,34 @@
...
+    "example": "ACC-1",
+    "github_request_issue": 1454,
+    "homepage": "https://celldive.dsmz.de/",
+    "mappings": {
+      "cellosaurus": "DSMZCellDive"
+    },
+    "name": "CellDive",
+    "pattern": "^ACC-\\d+$",
+    "synonyms": [
+      "DSMZCellDive"
```
cthoyt, line 14114: please maintain the old prefix dsmz as a synonym here

src/bioregistry/data/bioregistry.json
```diff
@@ -30676,55 +30705,6 @@
       "prefix": "P1930"
     }
   },
-  "dsmz": {
-    "cellosaurus": {
```
cthoyt, line 30680: it looks like a lot of components got lost, please make sure they are restored. How did you do the rename exactly? When I have done this, I just change the prefix, add the old prefix to the synonym list, then run `bioregistry lint`

nagutm: I made the following changes when I renamed it:
- removed the `re3data` mapping which was not mapping to CellDive, but DSMZ in general
- removed `example_extras` because only the identifiers with ACC- format were for cell lines. For example, [DSM-2](https://www.dsmz.de/collection/catalogue/details/culture/DSM-2) is for Paenibacillus macquariensis subsp. macquariensis, [PV-0998](https://www.dsmz.de/collection/catalogue/details/culture/PV-0998) is for Apple chlorotic leaf spot virus ACLSV, etc. 
- Updated the regex pattern to only include ACC which is what covers cell lines. 

The other identifier formats that are not ACC are related to specific microorganisms which felt like it was out of the scope for the `celldive` prefix so I intentionally removed them. Should these be maintained under a separate prefix or is it better to maintain this as part of the `celldive` prefix?

cthoyt: ahh I see what is happening. so we should probably maintain a superset `dsmz` prefix that is a union of all of these different vocabularies, since in practice they are conflated all together (and the ACC-, PV-, etc. work as janky subspaces)

nagutm: In that case I can modify this PR to just be about adding the `celldive` prefix and create a new PR that modifies the existing `dsmz` prefix accordingly.

src/bioregistry/data/bioregistry.json
bgyori: In the latest iteration, `celldive` is curated as `part_of` `dsmz` while `dsmz` remains as its own entry so the synonym isn't needed, I think.

bgyori:
> In that case I can modify this PR to just be about adding the celldive prefix and create a new PR that modifies the existing dsmz prefix accordingly.

I actually think the right scope for this PR would include any changes to `dsmz` that are directly related to the addition of the new `celldive` prefix. Otherwise, we would end up with two mappings to `cellosaurus:DSMZCellDive`, one from the new `celldive` and one from the existing `dsmz` which doesn't make sense.

bgyori:
I updated the description for `celldive` to mention DSMZ explicitly, and added reviewer info. However, I think this needs a bit more investigation. Basically, there are multiple identifiers involved in CellDive, and the URL pattern on the dedicated CellDive website is not the same as the identifier that is resolved via the core DSMZ website. The PR as it currently stands is related to the `ACC-` identifiers that are DSMZ-related rather than what is called the CellDive ID listed in the table in the attached screenshot or the name appearing in the URL pattern.

image

Issue #187, opened 2021-09-29
Add prefix Orphanet


matentzn:
### Prefix

Orphanet

### Name

Orphanet

### Homepage

https://www.orpha.net

### Description

Orphanet is a unique resource, gathering and improving knowledge on rare diseases so as to improve the diagnosis, care and treatment of patients with rare diseases.

### Example Identifier

http://www.orpha.net/ORDO/Orphanet_

### Regular Expression Pattern

[0-9]+

### Contributor ORCID

https://orcid.org/0000-0002-7356-1779

### Additional Comments

There is already a lowercase `orphanet` namespace but it uses the wrong URL prefix as well as we prefer this to be upper case due to convention.

cthoyt:
The Bioregistry already has two related prefixes, that are not the same:

1. http://bioregistry.io/registry/orphanet which is the main orphanet vocabulary
2. http://bioregistry.io/registry/orphanet.ordo which is the Orphanet Rare Disease Ontology 

`ordo` is already tagged as a synonym of `orphanet.ordo`. 

It looks like this request is mixing parts of each, using the prefix from the main vocabulary but the provider format URL for the ORDO vocabulary. I'm not really sure where the confusion comes from, but maybe it's a systematic mistake propagated from the Biolink Model's Biocontext (see http://bioregistry.io/registry/orphanet.ordo). I manually curated the mappings from that resource and noted that they were mix and matching before

matentzn:
I dont so much care about the prefix here, but the URL prefix needs to be added somewhere so I can run conversions correctly!

cthoyt:
I mentioned in that other issue (https://github.com/mapping-commons/sssom-py/issues/161) that I recently actually added this URI prefix in biopragmatics/bioregistry@21906c3. Is that sufficient?

matentzn:
It is! Thats the important bit.

dhimmel:
> The Bioregistry already has two related prefixes, that are not the same

I'm confused about the difference between ORPHA and ORDO terms, so coming here for help. And possibly finding some disambiguating description that we can add to Bioregistry to help users.


[...]

PR #1493, opened 2025-04-01
Collapse `orphanet.ordo` into `orphanet`

cthoyt:
Follow-up to #187. Supersedes #1003.

This PR does the following:

1. Makes sure that the URI format for both `orphanet` and `orphanet.ordo` is `http://www.orpha.net/ORDO/Orphanet_$1`
2. Set the "has_canonical" property for `orphanet.ordo` to point to `orphanet`. This makes `orphanet` the "preferred" prefix, rather than deleting `orphanet.ordo` or merging it into `orphanet` directly
3. Updates the OBO Context to map from `orphanet` instead of `orphanet.ordo` to the context's preferred `Orphanet`. Therefore, @matentzn and @anitacaron need to approve of this. It also includes the results from re-generating the OBO export artifacts

This is a bit of a trickier situation than other merges, since `orphanet` and `orphanet.ordo` have been duplicated in many external registries. Therefore, it's useful to see comments at https://github.com/biopragmatics/bioregistry/issues/187#issuecomment-1706197187 to justify doing this.

I also considered collapsing in a different way in https://github.com/biopragmatics/bioregistry/tree/orphanet-collapse-2, where I deleted `orphanet.ordo` completely. This seems like a worse solution because of the duplication


cthoyt:
@matentzn @anitacaron ping

reminder that when we wrote the governance for prescriptive contexts, we left a 4 week window for the contexts' managers to give feedback (after which I will merge this anyway)

cthoyt:
@nagutm @anitacaron this will get merged on April 29 if there are no comments

matentzn (approved):
Yes agreed, this is the practical way of doing it.. Any other solution would have to involve some active commitments from Orphanet, this is the most sane one without.

PR #1494, opened 2025-04-02
Curate Mappings With 0.5-0.6 Similarity Score

nalikapalayoor:
This PR updates curated_mappings.sssom.tsv with mappings that had similarity score between 0.5 and 0.6.


nalikapalayoor:
I closed this PR because there was an issue with the way I made the branch. The correct PR for this is #1496

PR #1495, opened 2025-04-02
Update ISBN resolver and add examples

bgyori:
This PR addresses issues raised in https://github.com/EnvironmentOntology/envo/issues/1619 and https://github.com/protegeproject/protege/issues/1271#issuecomment-2773592459 related to the previous default resolver for `isbn`: isbndb.com which doesn't support all ISBN syntax. I added a `uri_format` using isbnsearch.org to serve as a default overriding the imported resolvers since isbnsearch.org supports resolving various flavors of valid ISBN IDs.


src/bioregistry/data/bioregistry.json
```diff
@@ -60327,7 +60331,8 @@
     "synonyms": [
       "ISBN-10",
       "ISBN-13"
-    ]
+    ],
+    "uri_format": "https://isbnsearch.org/isbn/$1"
```
cmungall, line 60335: is this the PURL?

Are we not conflating URLs for humans and identifiers again?

bgyori: I don't think there is an issue here: this is simply a preferred override over the implicit imported resolver from identifiers.org (which is isbndb.com and is still there as one of the resolvers, just not the preferred one) which works better. Both of these are third-party providers of ISBN information, and I don't think either would be considered a "PURL". Is there a PURL for ISBN that should be added?

PR #1539, opened 2025-04-25
Merge `jax` into `jaxmice`

### LINKED ISSUE #1533: Clarify `jax` vs `jaxmice`
While reviewing metaregistry mappings with @nagutm, we noticed that there are two separate records for [`jax`](https://bioregistry.io/jax) and [`jaxmice`](https://bioregistry.io/jaxmice) but they ultimately both represent mouse strain identifiers and resolve to the same set of pages (unless we are missing something). This also affects how meta-registry mappings are interpreted and how information propagates into the combined Bioregistry record where there are a few suspicious things such as "Information about the C57BL/6J. Includes genetic background and disease data." being the description for `jax`.

bgyori:
This PR merges the `jax` prefix into `jaxmice`:
- We take the union of all mappings, except Biocontext whose `JAX` entry is preserved as a mapping and `JAXMICE` is dropped.
- Retained `jax` as a prefix synonym
- Updated website and description for clarity

Fixes #1533


cthoyt:
thanks!

PR #1545, opened 2025-04-26
Add preliminary Sage Bionetworks collection

allaway:
This is an initial collection of resources that we'll be linking to thru the bioregistry resolver from Synapse.


cthoyt:
Thanks @allaway - there are three updates to make this pass CI:

1. `jax` was renamed to `jaxmice` in #1539, so it should be replaced
2.  `metacyc` is not a standard prefix. Do you mean [`metacyc.compound`](https://bioregistry.io/metacyc.compound)?
3.  By `sabiork`, do you want to refer to [`sabiork.reaction`](https://bioregistry.io/sabiork.reaction)? Or to [`ec`](https://bioregistry.io/ec)?

allaway:
I have some check on our side for valid prefixes but I think I misunderstood the API and am treating synonyms as valid prefixes. Oops. Should be fixed now! Thanks!

PR #1547, opened 2025-04-29
New provider for `pdb`: `rnaprodb`

nalikapalayoor:
This PR curates a new prefix from RNA Protein Database
Website: https://rnaprodb.usc.edu/
Pubmed: https://pubmed.ncbi.nlm.nih.gov/40126909/


src/bioregistry/data/bioregistry.json
```diff
@@ -102567,6 +102567,34 @@
...
+      "name": "Remo Rohs",
+      "orcid": "0000-0003-1752-1884"
+    },
+    "contributor": {
+      "email": "palayoor.n@northeastern.edu",
+      "github": "nalikapalayoor",
+      "name": "Nalika Palayoor",
+      "orcid": "0009-0008-8406-631X"
+    },
+    "description": "RNAproDB is an interactive dashboard designed to assist research in structural analysis of RNA-protein conplexes. For each complex, the dashboard gives information about the nucleic acid structure as well as interacting proteins.",
```
cthoyt, line 102582: It's not clear from this description what the identifiers in this resource are. Please update the description to first describe what entities this semantic space is about, then after, it's okay to give a more general description of the resource.

Please also take a look at the guidelines on writing a good description: https://github.com/biopragmatics/bioregistry/blob/main/docs/CONTRIBUTING.md#writing-a-good-description

cthoyt (changes requested):
after looking at https://rnaprodb.usc.edu/1a1t, this appears to be a provider for PDB (see corresponding record at https://www.wwpdb.org/pdb?id=1a1t)

can you please adjust the curation accordingly to be a provider within the `pdb` record?

bgyori:
I believe the identifier here is actually a PDB identifier. Does this resource have its own identifiers or is it a provider for PDB?

nalikapalayoor:
Yes after looking again I agree this is a provider for PDB, I'll update this curation with that.

PR #1549, opened 2025-05-01
New provider: cyfidb

nalikapalayoor:
CyFidb is is a resource related to cyctic fibrosis. It allows searching from 67 variants and more than 100 types of cells and tissues used in research. The database also references the studies that each part of the database came from. It uses GenBank, Ensembl, Uniprot, and HGNC identifiers for each entry. This PR adds cyfidb as a provider for GenBank, Ensembl, and HGNC. It was unclear to me which branch of Uniprot this resource should go under, further discussion there would be helpful. 

Pubmed: https://pubmed.ncbi.nlm.nih.gov/40148144/
Website: https://cyfidb.di.fc.ul.pt/


bgyori (changes requested):
The `uri_format`-s don't make sense in these entries: these appear to be URIs from the original resources where these identifiers are defined, rather than URIs from CyFidb, i.e., the provider being defined here. Does CyFidb actually resolve these IDs? If not then we don't need to add it as a provider.

nalikapalayoor:
Ok I see. No it doesn't look like CyFidb resolves the IDs, it is just using the IDs from the original resources. I think I misunderstood what qualifies something to be curated as a provider, thank you for clearing that up!

bgyori:
Okay, I'll close this then.

Issue #1212, opened 2024-10-18
Proposal for Bioregistry SOP: "No manual registration of OBO URI prefixes"


matentzn:
Some [context](https://github.com/biopragmatics/bioregistry/pull/1175#issuecomment-2417053823).

I would like to suggest the following addition to the bioregistry SOP:

> OBO URI prefixes such as `http://purl.obolibrary.org/obo/ABC_` are automatically synchronised with the [OBO Foundry metadata](https://obofoundry.org/registry/ontologies.yml) and [prefixes](https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/registry/obo_context.jsonld). To avoid misusing bioregistries community driven editorial workflow to squat OBO PURLs for ontologies that are not officially accepted in OBO, we disallow manual curation of OBO URI prefixes in favour of full automatic synchronisation with the OBO prefix file.

@nataled that is what I meant - if OBCI say is accepted in OBO, the OBO URI prefix will automatically finds its way into bioregistry through the OBO metadata.

nataled:
@matentzn I understood. We'll need a suitable revision to the Foundry NOR instructions to account for this: In addition to checking the Bioregistry and BioPortal, submitters will need to also check (at least) the NOR Dashboard (for ontologies in the in-between).

cthoyt:
This should be more generic to say that new resources should not park on external IRI spaces. This isn't only applicable for OBO Foundry, but also the Crop Ontology IRI space, Identifiers.org IRI space, Name-to-Thing IRI space, RRID IRI space, or anything similar. For people trying to pick something that is PURL-like, they could consider w3id.org or similar

We can't make a technical solution that says non-OBO prefixes can't use OBO IRI space, as I've linked a few times in this discussion, there are people doing this whether we like it or not, [for a variety of reasons](https://cthoyt.com/2023/03/11/obolibrary-masquerade.html). Therefore, I propose that what you're asking for is articulated in terms of improving reviewer guidelines. How can you educate Bioregistry reviewers to spot these situations, and help them to explain to submitters why they _shouldn't_ be doing this? Will one of you take the lead on writing a more coherent explanation?

matentzn:
I framed it a bit differently, lets work on this and see what you think: https://docs.google.com/document/d/1aIP8YjyVnMzKnksMbBu1kYriV9jlMejWDrZ7E3GmXlY/edit?usp=sharing

PR #1551, opened 2025-05-05
New prefix: CLM

cthoyt:


cthoyt:
cc @aleixpuigb @dosumis

bgyori:
I am missing some context for where this comes from but is there any concern that this addition conflicts with the ideas at https://github.com/biopragmatics/bioregistry/issues/1212?

cthoyt:
No, #1212 is only a discussion. I don't think that it's possible for the Bioregistry to support stopping people from using specific PURLs (e.g., OBO PURLs), and if we were pedantic about not accepting new curations including unregistered OBO PURLs, we wouldn't be able to represent several important resources

PR #1558, opened 2025-05-05
New provider: oprlm

nalikapalayoor:
This PR curates a new provider for Protein Data Bank. OPRLM uses PDB identifiers in its database. 

Pubmed: https://pubmed.ncbi.nlm.nih.gov/40133776/
Website: https://oprlm.org/


bgyori (changes requested):
I am not sure this is right. When I look at e.g., https://oprlm.org/proteins/652 or https://oprlm.org/proteins/935, the identifiers in the URL are not PDB identifiers. The PDB identifiers appear in the table on the page. So this is not a provider for PDB IDs. In fact, it looks to me like this would be a new prefix since these identifiers (i.e., 652 and 935) are introduced by OPRLM itself.

nalikapalayoor:
Yes I agree with you, I wil update this to curating it as a new prefix.

PR #1559, opened 2025-05-05
Update synapse record

### LINKED ISSUE #1538: Update Synapse Data Repository
### Prefix

synapse

### Explanation

Hello! Would it be possible to change the prefix to `syn` for consistency with synapse IDs and have `synapse` as a synonym? 
I think we might also want to change the endpoint that bioregistry resolves to - i.e. synapse.org instead of the DRS service - but I need to talk with our tech team first. 

Thank you!

### Contributor ORCID

0000-0003-3573-3565

bgyori:
This PR updates the `synapse` record based on the discussion at #1538:
- Added `syn` as a banana to improve the handling of identifiers
- Also added `syn` as a synonym for the prefix
- Curated a couple of mapping mismatches to avoid spurious alignment on `syn`
- Updated contact info
- In https://github.com/biopragmatics/bioregistry/issues/1538#issuecomment-2846690700, various new mappings were proposed, however, the `fairsharing`, `rrid` and `wikidata` entries there were all to subsets of these resources that we don't map to, generally because of these entries representing databases rather than identifier spaces, so I didn't add these.
- Updated URI format to the landing page for the given project

Fixes #1538


bgyori:
@susheel and @allaway, here are the proposed changes. @kevinboske you are the proposed as the contact person, is there a dedicated email address for you that we could add to the record? Thanks all!

bgyori:
It looks like `test_url_banana` is failing (this tries to check that the URL pattern with banana included can be resolved via identifiers.org) because identifiers.org can only resolve https://identifiers.org/synapse:41455251.1 and not https://identifiers.org/synapse:syn41455251.1. This is indeed an issue in that by default our logic is to add the banana to generated identifiers.org links... trying to figure out how to resolve this

```
def test_url_banana(self):
        """Test that entries curated with a new banana are resolved properly."""
        for prefix, entry in self.entries.items():
            banana = entry.get_banana()
            if banana is None:
                continue
            if prefix in IDOT_BROKEN:
                continue  # identifiers.org is broken for these prefixes
            example = bioregistry.get_example(prefix)
            self.assertIsNotNone(example)
            with self.subTest(prefix=prefix, banana=banana, peel=entry.get_banana_peel()):
>               self.assert_url(prefix, example)

tests/test_identifiers_org.py:91: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/test_identifiers_org.py:[98](https://github.com/biopragmatics/bioregistry/actions/runs/14863074366/job/41734699596#step:4:99): in assert_url
    self.assertEqual(302, res.status_code, msg=f"failed with URL: {url}")
E   AssertionError: 302 != 404 : failed with URL: https://identifiers.org/synapse:syn41455251.1
```

cthoyt:
We should defer changing the logic for bananas to a PR that doesn't have any data changes. This needs some more thought, since when a banana is present for any record that maps to identifiers.org, it's assumed that it is necessary for constructing a CURIE valid for resolution for identifiers.org.

allaway:
So sorry for the delay on this, life got ahead of me! I can confirm that the email in the PR is appropriate for Kevin Boske, and that the other changes look great. Thanks @bgyori and @cthoyt!

allaway-test:
Sounds good to me. Thanks @cthoyt !

PR #1560, opened 2025-05-05
New prefix: hissta

nalikapalayoor:
This PR curates a new prefix hissta. This database contains information about tissue samples from various tissues using different sampling techniques. 

Pubmed: https://pubmed.ncbi.nlm.nih.gov/40163697/
Website: https://kbds.re.kr/hissta/


bgyori:
This looks good but there is an issue with how identifiers are defined: the number used in the URL pattern, which is interpreted as a "sample index" is not the same as the "Sample ID" that is listed under the given record. For instance, the "sample index" is `72` in the example below but the"Sample ID" is `XEN046`. I couldn't find a way to resolve the Sample IDs

image

nalikapalayoor:
> This looks good but there is an issue with how identifiers are defined: the number used in the URL pattern, which is interpreted as a "sample index" is not the same as the "Sample ID" that is listed under the given record. For instance, the "sample index" is `72` in the example below but the"Sample ID" is `XEN046`. I couldn't find a way to resolve the Sample IDs

I didn't notice that, do you recomend we curate this as "unclear" instead and write a small comment about this? I agree I don't see a good way to resolve the sample IDs either.

bgyori:
> > This looks good but there is an issue with how identifiers are defined: the number used in the URL pattern, which is interpreted as a "sample index" is not the same as the "Sample ID" that is listed under the given record. For instance, the "sample index" is `72` in the example below but the"Sample ID" is `XEN046`. I couldn't find a way to resolve the Sample IDs
> 
> I didn't notice that, do you recomend we curate this as "unclear" instead and write a small comment about this? I agree I don't see a good way to resolve the sample IDs either.

Yes, let's do that

PR #1564, opened 2025-05-18
Fix metadata for biocyc

nalikapalayoor:
This PR fixes the metadata for the publication https://bioregistry.io/pubmed:29447345 under the biocyc prefix.


src/bioregistry/data/bioregistry.json
```diff
@@ -8510,7 +8510,10 @@
       "publications": [
         {
           "doi": "10.1093/bib/bbx085",
-          "pubmed": "29447345"
+          "pmc": " PMC6781571",
+          "pubmed": "29447345",
+          "title": "The BioCyc collection of microbial genomes and metabolic pathways",
+          "year": 2019
```
bgyori, line 8516: We don't want to edit directly into the `uniprot` key. To complete the metadata, we want to add a publication directly under the `publications` key. Also, there is an extra leading space in the `pmc` value.

PR #1566, opened 2025-05-19
There are medgen ids with 3 digits in Medgen

matentzn:
Example: https://www.ncbi.nlm.nih.gov/medgen/?term=Inherited+bleeding+disorder%2C+platelet-type


bgyori:
Thanks @matentzn I agree, it looks like `610` is a legit MedGen ID that can be resolved by the default `uri_pattern` https://www.ncbi.nlm.nih.gov/medgen/610. The current identifier pattern is adopted from mapped external sources, so we can override it with a change to allow 3-dipgit IDs.

src/bioregistry/data/bioregistry.json
```diff
@@ -68080,7 +68080,7 @@
         "Phenotype"
       ],
       "name": "MedGen",
-      "pattern": "^(?CN?\\d{4,7})$",
+      "pattern": "^(?CN?\\d{3,7})$",
```
bgyori, line 68083: This is within the `togoid` mapped block. I will move it to outside the block so it serves as a Bioregistry override.

bgyori: @matentzn small issue... the PR is from your `main` branch so I can't push to it. Would you mind reverting the change above inside the `togoid` block and instead adding a new key directly under the `medgen` key
```json
    "pattern": "^[CN]*\\d{3,7}$"
```

PR #1567, opened 2025-05-19
Update MedGen

matentzn:
@bgyori I think there was already a pattern there, and I did not _fully_ update it the way you wanted, but just did the small edit here.

This PR makes the following updates:

- Fixes handling of `CN`-containing identifiers
- Enables IDs with 2 and 3 digits

Supports https://github.com/monarch-initiative/mondo/pull/9073


bgyori:
Thanks! Just to explain my change: as I mentioned on the original PR, we can update the pattern as a Bioregistry override (i.e., a `pattern` directly under the `medgen` entry) rather than editing into a mapped external resource's data block (in this case, your change was within the `togoid` block) since these are automatically pulled in and we don't manually curate them here. I also added explicit examples to make sure we cover multiple ID lengths for testing purposes.

However, now that we're looking at this, I have become suspicious more generally about these IDs. MedGen has its own UIDs for entries and also lists UMLS CUIs alongside these IDs. To me it is suspicious that the C/D prefix allowed in the current pattern actually refers to UMLS CUIs and not MedGen UIDs. To investigate further, I wrote the following code:

```python
from collections import Counter
import pandas as pd

df = pd.read_csv('https://ftp.ncbi.nlm.nih.gov/pub/medgen/MedGenIDMappings.txt.gz', sep='|')
df = df[df['source'] == 'MedGen']
medgen_ids = set(df['source_id'])

Counter(len(medgen_id) for medgen_id in medgen_ids)
# Yields Counter({6: 123477, 7: 58434, 5: 6727, 4: 1549, 3: 139, 2: 13})

any(not medgen_id.isdigit() for medgen_id in medgen_ids)
# Yields False
```

So MedGen UIDs appear to be numbers consisting of 2-7 digits, i.e., an appropriate pattern would be something like`^\\d{2,7}$`.

In addition, we could curate `medgen` as a provider for UMLS CUIs.

bgyori:
Found more context in: https://ftp.ncbi.nlm.nih.gov/pub/medgen/README.txt

```
Note that MedGen anchors information to the value known as CUI, so most of these 
reports include the defining CUI. When the value of CUI is the letter C followed by integers,
the value corresponds to a record from Unified Medical Language System (UMLS). When the value
of CUI are the letters CN followed by integers, the value was created by NCBI,
because a corresponding concept from UMLS was not identified.

MedGen assigns an internal unique identifier to each record, the UID. The web site reports the 
MedGen UID.
```

cthoyt:
Thanks @matentzn and @bgyori. I updated the pattern and examples to be more explicit about what IDs are UMLS and MedGen.

bgyori:
Just as a follow up: the "CN-containing identifiers" are part of a parallel identifier standard introduced by MedGen that is a placeholder for non-existent UMLS CUIs (it is replaced by a UMLS CUI if one is created). So these are not the same as the MedGen UIDs.

Here is a specific example

image

where there is a 7-digit MedGen UID 1610338 as well as a CN-containing CUI: CN970821. Is the merged definition which treats CN as an optional starting string the right way to represent this?

cthoyt:
Hmm, this will need some further investigating. likely solution will be to split `medgen` into `medgen.uid` and `medgen.cid` to cover the two specific cases

matentzn:
> Hmm, this will need some further investigating. likely solution will be to split medgen into medgen.uid and medgen.cid to cover the two specific cases

Back story:

Medgen uids are primary Medgen identifiers; the ids that contain CN are basically _temporary UMLS terms that do not exist yet_. @kanems would be the very response and amazing community person to own the medgen ID space and can provide clarity.

kanems:
I think Nico summed things up pretty well, just one point to add is that **ALL** MedGen records have a UID, but _not all_ records have a MedGen CN-type ID (that field can be either CN-ID from MedGen _or_ CUI from UMLS ). 
So if you want one reliable, consistent MedGen ID: the UID is the one to use.

cthoyt:
@kanems it's great to have a person to go with some of the NLM activities! which prefixes can I tag you as the contact person? MedGen, UMLS, NCIT, etc?

PR #1575, opened 2025-05-22
Filter out URIs constructed from non-W3C-compliant prefixes

### LINKED ISSUE #1338: Prefix with space character?

cthoyt:
Closes #1338 (better this time) by filtering out non-W3C-compliant prefix synonyms that are used to construct URIs (such as bioregistry.io URIs). Notably, this removes URIs with prefixes that have spaces, slashes, and non-ascii characters like letters w/ umlauts

@matentzn


exports/contexts/obo.epm.json
```diff
@@ -26587,7 +26532,6 @@
             "http://bioregistry.io/string:",
             "http://bioregistry.io/swiss-model:",
             "http://bioregistry.io/topfind:",
-            "http://bioregistry.io/uniprot/swiss-prot:",
```
matentzn, line 26590: I assume this is removed because of another reason!

matentzn:
Aweeeeeesome!!! THANKS!

exports/contexts/obo.epm.json
cthoyt: the prefix synonym `uniprot/swiss-prot` is not W3C-compliant since it has a slash in it, so it got removed

PR #1617, opened 2025-07-07
Add new prefix: chemfont.chemical and chemfont.ontology

bgyori:
This PR curates ChemFOnt (https://www.chemfont.ca/) which came up in a discussion and I noticed it is not yet in Bioregistry. There are several issues I encountered when attempting to curate this resource:
1. There are three types of identifiers in this resource:
- Chemicals, e.g., `CFc000239295` which can be resolved directly as e.g., https://www.chemfont.ca/chemicals/CFc000239295. I curated this as the `chemfont.chemical` prefix.
- Ontology terms that are _distinct from specific chemicals_, e.g., CFo000000018. These can be browsed via https://www.chemfont.ca/ontology_browse but I was not able to find a URL pattern to resolve these. I curated this semantic space as `chemfont.ontology`
- Definitions that appear alongside ontology terms and appear to be numbered corresponding to these, e.g., CFd000000018 for the term above, but again these do not appear to be resolvable directly, and their importance to be curated as a distinct prefix is unclear.
2. Mapping to other registries
- The only external registry where ChemFOnt appears is EcoPortal, under the prefix `CHEMFONT`. EcoPortal's representation, however, relates to the original resource in a complicated way. For example, terms like `CFo000000018` are catalogued not by their IDs but by their standard names, e.g., https://ecoportal.lifewatch.eu/ontologies/CHEMFONT?p=classes&conceptid=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FChemFont.owl%23IndustrialProcess 😭. It also looks like specific chemicals like CFc000239295 aren't in the EcoPortal instance of ChemFont at all. So it is not clear if/how the external prefix is mappable to our representation in Bioregistry.


cthoyt:
A few other observations:

- CFd000000018 can be resolved at https://www.chemfont.ca/ontology_browse/term_info/18, but this is problematic since the ID doesn't match to the URL
- The downloads page links to an ontology download artifact https://www.chemfont.ca/system/downloads/1.0/chemfont.owl.zip, this matches the problems described in EcoPortal. This is a candidate for re-implementation in PyOBO if/when it's useful to integrate this resource downstream
- There appear to be several additional vocabularies in the [category definitions download link](https://www.chemfont.ca/system/downloads/1.0/category_definitions.zip) on https://www.chemfont.ca/simple/download

PR #1623, opened 2025-07-21
Curate potentially relevant papers

maayamanoj466:
This PR adds new curations to the curated_papers.tsv file.


bgyori:
Thanks @maayamanoj! I reviewed these and agree on the choices here.

PR #1656, opened 2025-08-23
Add GotEnzyme as provider for KEGG Reaction

maayamanoj466:
This PR adds a new prefix, GotEnzymes, which is a browsable database of enzyme parameter predictions.


cthoyt (changes requested):
Taking a look at this now, https://metabolicatlas.org/gotenzymes/ appears to be a wrapper around data from KEGG. 

Screenshot 2025-09-08 at 17 35 18

The solution for this PR would be to make providers for various components in KEGG.

For example, the first one links to https://metabolicatlas.org/gotenzymes/reaction/R01677, where it shows that R01677 is a `kegg.reaction` ID. Each of the other examples also correspond to KEGG IDs. However, it might be the case that this resource mixes multiple distinct IDspaces

cthoyt:
See also:

- https://bioregistry.io/registry/metatlas.reaction
- https://bioregistry.io/registry/metatlas
- https://bioregistry.io/registry/metatlas.metabolite

PR #1659, opened 2025-08-25
Add CSSO and NVS

cthoyt:


src/bioregistry/data/bioregistry.json
```diff
@@ -24745,6 +24748,28 @@
     ],
     "uri_format": "http://purl.bioontology.org/ontology/CSP/$1"
   },
+  "csso": {
```
bgyori, line 24751: Is this a typo? I think this should be `ccso`

cthoyt: Yes, thanks for the catch. Fixed in https://github.com/biopragmatics/bioregistry/pull/1660

PR #1697, opened 2025-09-17
Split SMID record

cthoyt:
This PR was originally to update the SMID record based on references I found in ChEBI, but it actually is a bit more complicated:

- The `smid` page takes the integer one https://smid-db.org/smid/137
- The `detail` page takes the long one https://smid-db.org/detail/bhas%2338%0D/

ChEBI links to the second kind, for example [CHEBI:79239](https://www.ebi.ac.uk/chebi/searchId.do?chebiId=CHEBI:79239) links to http://smid-db.org/detail/bhas%2338%0D/

This is the same situation as the HGNC ID vs the HGNC symbol situation.


cthoyt:
cc @CarMoreno do you know where these references come from in ChEBI?

PR #1711, opened 2025-09-28
New provider: cedar

kanghosaeyo:


cthoyt (changes requested):
hi @kanghosaeyo, thanks for the PR. a few things you can do to improve your contribution:

1. Use the imperative voice in the title (as if you're yelling at the computer) "Add new provider cedar..."
2. Make the title more descriptive of what you actually did "Add provider cedar to Immune Epitope Database prefixes"
3. Add a description in the PR that says what you did like: 
   > I curated a new publication <title> (<pubmed ID>) which is about <describe CEDAR>. It is a provider for IEDB  + give context for why it basically duplicates the endpoints.
4. Add curator provenance to the providers (i'll add a suggetion on the diff directly)
5. In the description for the providers, you don't need to describe what the identifiers do, that's covered by the record. Please describe what is the resource that's doing the providing

bgyori:
I think this now addresses all the issues, thanks @kanghosaeyo!

PR #1713, opened 2025-10-01
Fix description of provider cedar under iedb prefixes

kanghosaeyo:
This pull request updates the provider description for CEDAR under the IEDB entry to properly describe CEDAR as the providing resource rather than describing what the identifiers represent. The description now clarifies that CEDAR is a cancer-focused interface providing specialized access to IEDB data with domain-specific analysis tools and workflows.


cthoyt:
thanks!

PR #1738, opened 2025-10-28
Update MIXS

cthoyt:
closes https://github.com/GenomicsStandardsConsortium/mixs/issues/1071

cc @turbomam


Woolly-at-EBI:
will mention that you have another one too.

PR #1739, opened 2025-10-31
Add new prefix: who.va

bgyori:
This PR adds a prefix for representing codes used in the WHO's Verbal Autopsy standard. Though there isn't a resolver for these IDs, they are defined in the documentation at https://www.who.int/publications/m/item/verbal-autopsy-standards-the-2016-who-verbal-autopsy-instrument and are produced by various software tools as output.


bgyori:
This comes up in the space of ICD codes and related cause of death codes, e.g., in these algorithms: https://github.com/verbal-autopsy-software/openVA

PR #1785, opened 2025-12-04
Implement SKOSMOS ingestion and add initial curations

cthoyt:
[SKOSMOS](https://skosmos.org) is a lightweight ontology and controlled vocabulary lookup service, like OLS, but for SKOS-flavored vocabularies. There are several SKOSMOS instances listed by [BARTOC](https://bartoc.org/registries) that contain relevant identifier resources to incorporate.

This PR adds a workflow from creating Bioregistry records from SKOSMOS that can be quickly curated and the missing bits filled in.

depends on:

- [x] #2065
- [x] #2066


cthoyt:
cc @eilmiv

PR #1804, opened 2026-01-13
Update homepage and URI format for prefix: streptomedb

kanghosaeyo:


bgyori (approved):
Tested and the new URL works well

PR #1850, opened 2026-03-17
UniProt importer: use `abbrev`, not `id` field

ialarmedalien:
The UniProt importer currently uses the `id` field from the UniProt database download list, but the `id` field is just an internal UniProt identifier and not used in any of the public-facing data (i.e. in the mappings to other databases).

This PR edits the parser so that it takes the `abbrev` field instead, which is where the db prefixes are held.

Added tests for the UniProt parser at the same time.


cthoyt:
I'm going to close this - if there's any feedback on the previous comment then we can discuss a different way forward

Issue #1838, opened 2026-03-09
Ingest UniProt database registry and prefixes


ialarmedalien:
# UniProt 

## Description

UniProt provides a [list of external databases that it links to](https://www.uniprot.org/database?query=*), along with information on how to create the appropriate URLs for entries.

## Prefixes

185 freshly-baked prefixes and attendant database information

I'm interested in getting the prefixes used by UniProt integrated into the Bioregistry. As mentioned above, UniProt provides a list of external databases that it links to, with info including

- prefix
- db name, description
- URL
- format of link for an individual entity

Some prefixes are already aligned, either with the Bioregistry preferred prefix or with a synonym, and some represent a subclass or superclass (e.g. UniProt uses `merops` to refer to `merops.entry`). Other entries are missing from Bioregistry -- e.g. PATRIC.

It would be great to be able to incorporate this prefix data into Bioregistry programmatically. A coworker and I have already done some work on mapping the UniProt and Bioregistry prefixes where there is not an exact match.

What would be the best way to proceed?

bgyori:
Thanks @ialarmedalien! Indeed, this UniProt registry is already integrated into the Bioregistry (https://bioregistry.io/metaregistry/uniprot) but as you noted, the alignment is not complete. 

You can see a summary of the alignment here: https://bioregistry.io/api/metaregistry/uniprot/mappings.json, this covers 95 mappings currently.

To add more to this, we could do some combination of the following:
- The code for UniProt registry processing and alignment is implemented here: https://github.com/biopragmatics/bioregistry/blob/main/src/bioregistry/external/uniprot/__init__.py so if there is any general/programmatic improvement to the alignment, it could be implemented here.
- For prefixes that already exist in the Bioregistry but the automated alignment didn't find the correct mapping, manual mappings can be curated here: https://github.com/biopragmatics/bioregistry/blob/main/src/bioregistry/data/curated_mappings.sssom.tsv. You can see several examples of mappings to the UniProt registry curated here already.
- For prefixes that don't yet exist in the Bioregistry at all and are relevant, we might want to create new prefixes and then add the mapping (or let the auto-alignment create the mapping). Not sure how many of these there are, depending on the number, we could do this manually or find some more scalable solution.

@cthoyt might also have more sugesstions!

ialarmedalien:
I am pulling in a load of data from UniProt and want to remap prefixes from what they publish to what is registered in the Bioregistry (as far as is possible). I have various comments and questions.

[...]

PR #1852, opened 2026-03-18
Add missing short name to UniProt

cthoyt:
Related to https://github.com/biopragmatics/bioregistry/issues/1838

This PR fixes a bug in UniProt importer. It was previously using the wrong key `abbreviation` instead of `short_names` and at the same time, wasn't using strict parsing, so this wasn't caught.

### DIFF
--- a/src/bioregistry/external/uniprot/__init__.py
+++ b/src/bioregistry/external/uniprot/__init__.py
@@ -8,7 +8,7 @@ from collections.abc import Sequence
 from pathlib import Path
 from typing import Any, ClassVar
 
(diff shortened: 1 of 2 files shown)


cthoyt:
cc @ialarmedalien as I am working through your use case, I found this non-controversial issue which was a bit obfuscated by the other PR

ialarmedalien:
@cthoyt Thanks!

I wrote pretty comprehensive tests for the uniprot parser as part of my PR -- might be useful to add them to the repo regardless of what resolution is arrived at for the main issue. It should not take long to adapt them to the current parser and data structures.

PR #1889, opened 2026-04-10
Add prefix: g2p

hetvikhamar:
This PR adds Gene2Phenotype, which is a project that provides evidence-based models to help interpret genomic data. Its goal is to accelerate the diagnosis of children with developmental disorders. It can be used to effectively filter genotype data from the exome or genome wide sequencing of people with genetic disorders to identify variants and accelerate diagnosis.

### DIFF
--- a/src/bioregistry/data/collections.json
+++ b/src/bioregistry/data/collections.json
@@ -1717,7 +1717,7 @@
           "wikidata": "Q139055838"
         }
       ],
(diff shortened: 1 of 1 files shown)


bgyori:
This looks like a great addition, in fact I'm surprised this wasn't catalogued before since it's not a completely new resource. The `G2P\d{5}` identifier pattern for gene-phenotype records curated here appears to be novel and I couldn't find it registered in any of the external registries.

I made a couple of changes to improve the entry, otherwise looks good to me.

PR #1891, opened 2026-04-10
Add DataPLANT maintainers

cthoyt:
Hi @kdumschott @AngelaKranz, we're porting the Section Metadata WG Ontology Harmonization survey of ontology usage for DataPLANT into this repository. Can we keep listing you as contacts/responsible people in the consortium for the ontology usage list? We can add more / change this over time as well.

Here's how your consortium's list looks: https://bioregistry.io/collection/0000023

### DIFF
--- a/src/bioregistry/data/collections.json
+++ b/src/bioregistry/data/collections.json
@@ -1279,6 +1279,12 @@
           "github": "AngelaKranz",
           "name": "Angela Kranz",
           "orcid": "0000-0002-8000-0400"
+        },
+        {
+          "email": "h.doerpholz@fz-juelich.de",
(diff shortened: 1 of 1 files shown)


cthoyt:
@Hannah-Doerpholz you are also relevant here!

kdumschott:
Hi Charlie, that is fine. Myself or Angela are the main responsible people, with Hannah being more of an alternate (not sure if this is something you're clarifying in the list).

cthoyt:
@kdumschott thanks for the comment! right now we're not capturing roles (besides what the order suggests). if we update the data model, we'll capture this later :)

PR #1916, opened 2026-04-16
Add prefix: prospero

hetvikhamar:
Propspero is an international registry that contains records of over 435000 prospectively registered systematic reviews with health related outcomes. It creates publicly available registration records and issues a unique registration number for each registered review protocol.


bgyori:
This is a good entry though I had to make a lot of changes:
- Weirdly, the publication and contact were irrelevant to the actual resource, I did some research and found appropriate ones
- The pattern was incorrect in terms of the number of digits
- Set the name to PROSPERO which is the actual name of the resource, and updated the description with what it is

PR #1920, opened 2026-04-17
Add prefix: pneumowiki

hetvikhamar:
Pneumowiki is a database that contains information for the pathogen Streptococcus pneumoniae. It can be used to search for tags, genes, locus or protein names of S. pneumoniae. It includes data for 43 different strains.


bgyori:
This looks like a good addition. The pangenome identifiers introduced by PneumoWiki (`PNEUPAN\d{9}`) appear to be novel. It is worth noting that there are also many other identifier patterns one can resolve via PneumoWiki like SP_RS03910, SPD_RS03780, EF3030_RS03780 which are strain-specific IDs. These can be resolved via BioCyc (e.g., https://biocyc.org/gene?orgid=GCF_004291255&id=EF3030_RS03780) as well. Perhaps its easiest to just set these aside and focus on the genuine new prefix for the pangenome-based IDs.

PR #1921, opened 2026-04-19
Add prefix: preppi

hetvikhamar:
PrePPI is a database of predicted protein protein interactions. It contains about 1.3 million predicted interactions of the human proteome. These predicted interactions are determined through a Bayesian framework that combines structural, functional, evolutionary and expression information.


bgyori (changes requested):
This is a relevant resource but doesn't introduce novel identifiers: it resolves UniProt identifiers. We should curate this as a `provider` under the `uniprot` entry instead.

bgyori (approved):

PR #1938, opened 2026-04-29
Update SI Reference Point

Zack-83:
SI/SIRP was added with #1901 .

This PR adds a contact (https://github.com/Crparedes) and adjusts example and definitions


cthoyt:
Great, thanks @Zack-83! The only thing missing is to add your contributor information under the `contributor_extras` list in each of these records. I can take care of that

PR #1943, opened 2026-05-05
Fix UniProt regular expression

cthoyt:
Identifiers.org appeared to have added many extra parts to the uniprot namespace to cover uniprot.chain and uniprot.isoform - these are explicitly different semantic spaces, despite being "derived" from the parent protein's identifier, so they shouldn't validate under `uniprot`


bgyori:
I think we should do this since the identifiers.org pattern is not valid in this setting. Worth noting that UniProt also declares their own "official" pattern at https://www.uniprot.org/help/accession_numbers which is currently showing 
```
[OPQ][0-9][A-Z0-9]{3}[0-9]|[A-NR-Z][0-9]([A-Z][A-Z0-9]{2}[0-9]){1,2}
```
which I think we could adopt (adding ^ and $ which is not part of the original). The proposed pattern here has a few issues like `[A-Z, 0-9]` (has a literal comma and space character) should be `[A-Z0-9]`, and we'd want to add one more set of parentheses around the pattern before adding the ^ and $ around it.

PR #1946, opened 2026-05-08
Update PaNET

cthoyt:
cc @kara-mela @hgoerzig

this PR adds an explicit download URL for the PaNET OWL, which enables tools like sssom-curator to automatically find, index, and propose mappings for it

this PR also improves several other metadata in this ontology


kara-mela:
Hi @cthoyt, 

thanks for adding me to the contact people of PaNET. As I am currently working on a limited contract, I suggest to also add people with permanent positions, e.g. @paulmillar, @hgoerzig, @RKrahl.

Issue #1931, opened 2026-04-28
Sub-collections


cthoyt:
Define a notion of sub-collections. This might be something like "collection A imports collection B" and then all of its entries get propagated up

cc @StroemPhi

StroemPhi:
We might want to think a bit more about how we actually want this in the Semamntic Farm. My current use case would be to be able to differentiate the ontologies within a collection between (a brainsorming):

- [x] internal ontologies = created by the consortium of the collection. E.g. VIBSO  --> implemented already
- [ ] external but curated by the consortium of the collection, e.g. CHMO, RXNO, RSC curated by NFDI4Chem. (note from CTH: Could potentially implement this by crawling repos for each ontology, getting contributor lists, and mapping back to ORCiD)
- [ ] external/internal used in production within the services of the consortium of the collection, e.g. CHMO, RXNO, CHEMINF, CHEBI, nmrCV, BFO, PROV-O, DCAT, ... used in Chemotion and/or  Search Service
- [ ] external ones used for research purposes --> not in any associated services of the consortium of the collection, but indexed to compare with the ones used, or to learn from, or to maybe build mappings. E.g. FIX, REX and Gainsville Core, as they are considered outdated/orphaned, but still are necissary for mappings/integration or to salvage in terms of minting equivalent new terms in curated ones

cthoyt:
one technical solution for this would be to allow a collection to define a controlled vocabulary of (local) tags, that entries in the collection could annotate on.

that would be if we want to keep track of these manually.

I already demonstrated we could have an "algorithm" that determines what's first-party. Maybe we could establish other workflows for annotating entries in different ways

StroemPhi:
related to https://github.com/nfdi-de/section-metadata-wg-onto/issues/32 -> we might open another sub-issue of this epic in our WG Onto repo later for this. But keeping this here for now, seems best to me

PR #1958, opened 2026-05-17
Implement collection tags

### LINKED ISSUE #1957: Internal collection tags
> We might want to think a bit more about how we actually want this in the Semamntic Farm. My current use case would be to be able to differentiate the ontologies within a collection between (a brainsorming):
> 
> - [x] internal ontologies = created by the consortium of the collection. E.g. VIBSO  --> implemented already
> - [ ] external but curated by the consortium of the collection, e.g. CHMO, RXNO, RSC curated by NFDI4Chem
> - [ ] external/internal used in production within the services of the consortium of the collection, e.g. CHMO, RXNO, CHEMINF, CHEBI, nmrCV, BFO, PROV-O, DCAT, ... used in Chemotion and/or  Search Service
> - [ ] external ones used for research purposes --> not in any associated services of the consortium of the collection, but indexed to compare with the ones used, or to learn from, or to maybe build mappings. E.g. FIX, REX and Gainsville Core, as they are considered outdated/orphaned, but still are necissary for mappings/integration or to salvage in terms of minting equivalent new terms in curated ones 

 _Originally posted by @StroemPhi in [#1931](https://github.com/biopragmatics/bioregistry/issues/1931#issuecomment-4334388003)_

cthoyt:
Closes #1957 

This PR enables each collection to define local "tags". One or more tags can be applied to each entry in the collection to give context on to why the ontology was included, what's its purpose, etc.

Ideally, these tags won't be degenerate of already available information in the Bioregistry (e.g., don't use tags to say which entries are ontologies and which are databases)

There's also already an automated way of determining which resources are "first party", so that would be another anti-pattern for usage of tags.

Tags can be used in combination or in place of narrative descriptive text along with each entry, such as in the example below for NFDI4Chem's ontology collection. This PR uses the NFDI4Chem example from https://github.com/biopragmatics/bioregistry/issues/1931#issuecomment-4334388003 as a case scenario:

Screenshot 2026-05-17 at 13 00 14


src/bioregistry/data/collections.json
```diff
@@ -561,24 +561,56 @@
...
         "chiro",
-        "chmo",
+        {
+          "prefix": "chmo",
+          "tags": [
+            "collaborative",
+            "production"
+          ]
+        },
+        {
```
cthoyt, line 590: @StroemPhi note that I added DCAT to the NFDI4Chem collection here - this is new. All other changes are just adding context to existing collection entries

cthoyt:
@allaway this is also a feature that might be interesting for you to internally organize the Synapse collection

PR #1988, opened 2026-06-07
Update multiple ICD11 URI formats

bgyori:
Several ICD11-related `uri_format`-s are using hard coded specific release dates or simply outdated URLs. It makes sense to replace these with a generic `latest-release` while keeping old URI formats as legacy provider entries. I also finally found an (unofficial) endpoint to resolve `icd11.code` identifiers for which there doesn't appear to be an official endpoint despite widespread use.

Specific changes:
- Under the `icd11` prefix
  - Update the main `uri_format` and add the old one as a legacy
  - Update both `mms` and `foundation` provider `uri_format`-s and systematically add legacies for them using specific older dates
- Under the `icd11.code` prefix
  - Add a new working `uri_format`, previously missing


cthoyt (approved):
Thanks!

PR #1989, opened 2026-06-07
Update prefixes: `rapdb.locus` and `rapdb.transcript`

hetvikhamar:
rapdb is a database that aims to provide annotation of the rice reference genome. It was developed is 2004 after the completion of the Oryza sativa ssp. japonica cv. Nipponbare genome sequencing. Its main purpose is to give analyses of the rice structure and function.


bgyori:
Upon review, we already have a `rapdb.locus` prefix that some of this metadata should be added under. Also, most of the contact information is incorrect and has to be updated.

bgyori:
I updated `rapdb.locus` and `rapdb.transcript` to add contact info, override the outdated homepage and URL pattern, add the new publication, and keyworkds, as well as a standardized name.

PR #1996, opened 2026-06-15
Enable displaying tags as columns

cthoyt:
requested by @FederalGobbo 

This PR enables marking individual tags in a collection as whether they should be displayed as a badge (inline in the description) or as its own column


cthoyt:
I'm going to close this since the the NFDI4Immuno use case doesn't actually need this feature

PR #1997, opened 2026-06-22
Add prefix: `moonprot`

Thadeus0422:
This PR adds a new prefix for the MoonProt DB which introduces identifiers for proteins that perform multiple functions


bgyori:
Thanks @Thadeus0422, this looks good, I just made a few updates.

PR #2000, opened 2026-06-24
Add initial NFDI4Immuno collection

FederalGobbo:
All the currently in use ontologies of NFDI4Immuno have been entered here.

### DIFF
--- a/src/bioregistry/data/collections.json
+++ b/src/bioregistry/data/collections.json
@@ -2618,13 +2618,13 @@
       ],
       "resources": [
         {"prefix": "bioregistry", "tags": []},
-        {"prefix": "doid", "tags": ["disease"]},
+        {"comment":"Disease ontology used to describe patients disease / disease of research focus","prefix": "doid", "tags": ["disease"]},
         {"prefix": "hancestro", "tags": ["ancestry"]},
-        {"prefix": "snomedct", "tags": ["Ethnicity"]},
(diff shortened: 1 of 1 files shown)


FederalGobbo:
@cthoyt : Thank you for your patience. Can you please check if the alterations are acceptable? 
IT should only contain NFDI4Immuno additions in the way you requested. 
Best wishes, 
Sebastian

cthoyt:
@FederalGobbo @NFDI4Immuno-FLI given that you’ve defined categories that only get used once, I’m wondering if the overhead of using tags is worth it. Might a comment saying “this is what we recommend for entities of type X” be good enough?

FederalGobbo:
@cthoyt : I have adapted the code accordingly. Hope it meets your expectations. 
Best wishes, 
Sebastian

FederalGobbo:
@cthoyt : I start getting a bad conscience, as I seem to have caused more trouble then benefit. Let me know if I can help you any further.

cthoyt:
@FederalGobbo not at all! I just did a little reformatting using the automatic linter and then noticed that the names of some of the resources needed to be updated. Thank you for the submittion! I will ping you when it's live.

cthoyt:
@FederalGobbo @NFDI4Immuno-FLI updates are live at https://semantic.farm/collection/0000043!

FederalGobbo:
@cthoyt Thank you Charles!

PR #2001, opened 2026-06-25
Calculate resource inter-dependencies

cthoyt:
As part of https://github.com/nfdi-de/section-metadata-wg-onto/issues/85, this script parses all resources annotated in the Bioregistry with an ontology artifact to identify what other resources they use, then puts them in the `depends_on` field

at a high level, this will allow for seeing which upper-level ontologies each resource uses (if any)

more detailed rules about what is an actual "dependency" are probably needed. e.g., cross-references probably shouldn't count this way

I also need to decide if this data should really live in Bioregistry or somewhere else.

cc @StroemPhi @matentzn

### DIFF
--- a/src/bioregistry/curation/add_depends_on.py
+++ b/src/bioregistry/curation/add_depends_on.py
@@ -1,0 +1,164 @@
+"""Analyze."""
+
+import itertools as itt
+
+import bioregistry
+import click
+import rdflib
+from bioregistry import Resource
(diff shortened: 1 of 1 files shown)


matentzn:
I think this information would be very valuable but I worry about completeness, correctness, runtime, and complexity adding this to bioregistry built... I would say this is more like a research paper with a dashboard ala obo-health then an ETL pipeline folded into br.

PR #2002, opened 2026-06-25
Improve OBO Foundry contact information

cthoyt:
Based on https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2909

OBO Foundry is currently going through procedural discussions on if/when to incorporate this information, but Bioregistry's governance is much more aggressive towards curating information that can be publicly found by searching GitHub or in the author lists of papers


cthoyt:
follow-up idea, have a reconciliation system to push metadata from Bioregistry upstream (e.g., back to OBO Foundry), then deduplicate when the bioregistry annotations are the same as trusted upstream sources (e.g., OBO Foundry, not BioPortal since metadata there is low quality)

PR #2097, opened 2026-09-17
Change the GARD uri_format

matentzn:
GARD has changed the ID format to zero padded, and the old URI format does not work anymore. For context see this comment by the GARD contact Eric: https://github.com/monarch-initiative/mondo-ingest/issues/917#issuecomment-5607967165.

@cthoyt I dont know how to best document that GARD:0001111 and GARD:1111 is identical from the GARD standpoint. 

- ✅ https://rarediseases.info.nih.gov/diseases/6038/index
- ❌ https://rarediseases.info.nih.gov/diseases/0006038/index
- ✅ https://rarediseases.info.nih.gov/?gard_id=0006038, redirects to https://rarediseases.info.nih.gov/diseases/6038/chikungunya-fever
- ✅ https://rarediseases.info.nih.gov/?gard_id=6038, redirects to https://rarediseases.info.nih.gov/diseases/6038/chikungunya-fever

This is breaking Monarch, OLS and other linkout systems to GARD.

cc @ericsid.


cthoyt:
It's unfortunate that GARD changed its IDs (PDB did the same thing recently, despite my pleas against it), but I think that switching the URI format and adding the comment what happened is good enough. I'll update this record to keep the old URI as a "legacy" so anyone who has old URIs can still turn them into CURIEs

however, GARD could release a SSSOM file that has a mapping from old IDs to new ones to support people in transitioning

cthoyt (approved):

matentzn:
Thanks @cthoyt - looks great. @ericsid from your perspective, thinking now about the "string representation of your GARD identifiers for external consumption", is the situation

- only GARD:123 is supported moving forward
- only GARD:0000123 is supported moving forward
- both of the above are supported moving forward, but 
   - only one (which?) is preferred
   - you have not yet made up your mind which is preferred

ericsid:
@matentzn - Currently, the situation is more of a "both of the above are supported moving forward" by the GARD website at this time though, I tend to prefer the zero padded format. Please note that the GARD website is managed by a colleague (Meera Shah) though I continue to act as the owner for our rare disease list and GARD identifiers since I manage other efforts promoting interoperability and FAIR data practices with rare disease researchers. 

@cthoyt  - The original rationale behind this change was to better support use of GARD IDs in other applications beyond GARD's public health website so that we could foster interoperability in more NIH data science efforts and resources (e.g. https://raresource.nih.gov/literature/disease/0001608 ; https://rdas.ncats.nih.gov/disease?id=GARD:0001608 ; etc.). If you think It will be helpful for you or other knowledge providers to have an SSSOM file that addresses our legacy IDs, I can add that to our backlog - just let me know.

